PR #21, opened 2024-07-14
multi line interactive on {} working (a bit ugly)

### LINKED ISSUE #13: multiline support in stdin repl
right now when not passing in script files, so in repl mode, each line must stand by itself, that's bad

ldemailly:
fixes #13


lexer/lexer.go
```diff
@@ -7,15 +7,22 @@ import (
...
+	input    string
+	pos      int
+	lineMode bool
 }
 
+// Mode with input expected the be complete (multiline/file).
 func New(input string) *Lexer {
 	l := &Lexer{input: input}
 	return l
 }
```
ccoVeille, line 19:
```suggestion
 func New(input string) *Lexer {
 	return &Lexer{input: input}
 }
```

ccoVeille (commented):
LGTM 👍

Minor stylistic suggestions anyway

lexer/lexer.go
```diff
@@ -86,7 +93,11 @@ func (l *Lexer) NextToken() token.Token { //nolint:funlen // many cases to lex.
 	case '"':
 		return token.Token{Type: token.STRING, Literal: l.readString()}
 	case 0:
-		return token.Token{Type: token.EOF, Literal: ""}
+		if l.lineMode {
+			return token.Token{Type: token.EOL}
+		} else {
+			return token.Token{Type: token.EOF}
+		}
```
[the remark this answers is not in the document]
ldemailly: I think we went through that last time, when there is exactly 1 line and it looks symetrical, while I'm very anti else in general, I don't in this case :)

lexer/lexer.go
ldemailly: fixed in https://github.com/grol-io/grol/commit/5b28ec0cef422e46b5ba9867ffb7726e18b4c192
was meant to be an PR but I need to fix the permissions

lexer/lexer.go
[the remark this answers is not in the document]
ccoVeille: We might have talked about indeed.

But let's say you didn't convince me, and I didn't convince you either 😅🤭🤣

lexer/lexer.go
ccoVeille: > not sure it's how it is, must have had something else 

As far as I know, only simple types like int, string need a variable.

https://github.com/golang/go/issues/45624

PR #32, opened 2024-07-15
capture print/log when using EvalString (for the discord bot)

ldemailly:


eval/eval.go
```diff
@@ -13,11 +14,13 @@ import (
 )
 
 type State struct {
-	env *object.Environment
+	env   *object.Environment
+	Out   io.Writer
+	NoLog bool // turn log() into print() (for EvalString)
```
ccoVeille, line 19: I would have used a logger interface, then I would have define a "printer" when I needed it.

I'm unsure by the current implementation

ldemailly: So I'm hesitating between providing 3 result for "evalstring"

- the eval result
- the stdout equiv
- the stderr equiv

for now this just puts everything in a string, more or less how it appears:

Screenshot 2024-07-15 at 1 09 59 PM

vs what you see in repl_test.go
ahah! new file is missing... hold on

ldemailly: https://github.com/grol-io/grol/pull/32/files#diff-52699463a4b9bdc9d5b654f3aead9c1ef3d9c671f99c66c7970b0bba34979045R21-R26

So yes if I go for full split of 3 result, the logger would be pluggable
(the other issue is I'm still using my pre slog logger that... I happen to like but has a global config for output)

ccoVeille: I'm unsure what it may look alike right now. But I think it would be better to separate rendering from the current process

ldemailly: follow up in #33

ccoVeille: Thanks

ldemailly: See how it is used in https://github.com/grol-io/grol-discord-bot/pull/5

PR #35, opened 2024-07-15
Expose errors (for bot). 

ldemailly:
for https://github.com/grol-io/grol-discord-bot/issues/7

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -48,14 +48,15 @@ func EvalAll(s, macroState *eval.State, in io.Reader, out io.Writer, options Opt
 }
 
 // EvalString can be used from playground etc for single eval.
-func EvalString(what string) string {
+// returns the eval errors and an array of errors if any.
+func EvalString(what string) (string, []string) {
 	s := eval.NewState()
 	macroState := eval.NewState()
 	out := &strings.Builder{}
 	s.Out = out
 	s.NoLog = true
-	EvalOne(s, macroState, what, out, Options{All: true, ShowEval: true, NoColor: true})
-	return out.String()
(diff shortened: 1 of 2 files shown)


repl/repl.go
```diff
@@ -84,7 +85,8 @@ func Interactive(in io.Reader, out io.Writer, options Options) {
 }
 
 // Returns true in line mode if more should be fed to the parser.
-func EvalOne(s, macroState *eval.State, what string, out io.Writer, options Options) bool {
+// TODO: this one size fits 3 different calls (file, interactive, bot) is getting spaghetti.
```
ldemailly, line 88: @ccoVeille I realize ^

repl/repl.go
```diff
@@ -72,7 +73,7 @@ func Interactive(in io.Reader, out io.Writer, options Options) {
 			return
 		}
 		l := prev + scanner.Text()
-		contNeeded := EvalOne(s, macroState, l, out, options)
+		contNeeded, _ := EvalOne(s, macroState, l, out, options)
```
ccoVeille, line 76: Why ignoring errors here?

Maybe it's OK, but a comment would help

ldemailly: it's because they've been printed interactively, which is kinda why the spagheti of ifs and modes in EvalOne isn't great, I probably should just copy paste exactly for 3 modes, or ... clean up somehow

agreed on comment, adding

ldemailly: https://github.com/grol-io/grol/pull/37/commits/c09493cc02be18a857bf76c8ed1d51ada6561993

PR #51, opened 2024-07-20
trying full anchor for tag regexp even though...

### LINKED ISSUE #50: fix tag regex here and discord bot
thank I fixed that bug in fortio/workflows but when I copied pre workflow files from multicurl I forgot about, nice catch

_Originally posted by @ldemailly in https://github.com/grol-io/grol-discord-bot/pull/16#discussion_r1685219794_

ldemailly:
Fix #50


ldemailly:
@ccoVeille See commits and tags, tried everything, anchor don't work and yet what is right now does (and so did the version with `eeeeee` yet not -test... not sure why)

Issue #46, opened 2024-07-19
BUG: return ends the whole script instead of the current function


ldemailly:
```go
$ f=func() {return};f(); print("not reached")
nil
```

PR #52, opened 2024-07-20
introduce testscript, fix return bug, add map keys sorting, non color mode bug fix

ldemailly:
- [x] Introduce testscript
- [x] Failing test
- [x] Fixed bug reset color code being printed in non color mode
- [x] Sort object keys by type then content
- [x] Adding test of sample.gr's output
- [x] Fix bug #46


eval/eval.go
```diff
@@ -304,7 +304,7 @@ func (s *State) applyFunction(fn object.Object, args []object.Object) object.Obj
 	}
 	curState := s.env
 	s.env = nenv
-	res := s.evalInternal(function.Body)
+	res := s.Eval(function.Body) // Need to have the return value unwrapped. Bug #46
```
ccoVeille, line 307: Either it's a todo


```suggestion
	res := s.Eval(function.Body) // TODONeed to have the return value unwrapped. Bug #46
```

Either is was fixed and I would expect to do not mention "bug"

```suggestion
	res := s.Eval(function.Body) // Need to have the return value unwrapped. See #46
```

ldemailly: it _fixes_ Bug #46, I can add Fixes #46, there is no todo it's the fix

ldemailly (commented):
.

eval/eval.go
ldemailly: fixed in https://github.com/grol-io/grol/pull/53/commits/4097534d4e86fab40035cb38c8cf087d598da88d

ccoVeille:
LGTM, even if I'm reviewing once merged

PR #65, opened 2024-07-22
complete code and design revamp, optional () in if, -format option, idempotent, stdin input file option

### LINKED ISSUE #42: consider making `if` work without needing the ()

### LINKED ISSUE #57: parse(print(parse(grol code))) != parse(grol code)
ie parse .print doesn't always produce parseable code and it's not an identity function when re applied, it should

(then we get grol format too)

related to #38

ldemailly:
- [x] -format option
- [x] made me remove the extra () in expressions
- [x] which made the canonical if not have the () so we could run what we print - fixes #57 
- [x] which made we need to accept if without () - fixes #42 
- [x] Adding test for #57 based on the samples.
- [x] in progress complete revamp of token/lexer, interning everything possible
- [x] need to handle nested expressions (which do need ()) - ie pass the tests
- [x] print isn't pretty... (need to add tabs)


ldemailly:
it's a big too big to review, was almost more work to refactor than would have been to rewrite from scratch I think, and updating tests was somewhat painful, yet useful to have - mostly fyi @ccoVeille - I will merge but you can leave comments later and it's ok if you don't look at all

ldemailly:
token/ and lever/ though are I think interesting and reviewable

PR #73, opened 2024-07-28
Adding postfix expressions (++ and --) - changed evaluation order to be more human friendly (appears left to right now)

### LINKED ISSUE #17: add ++ and --
note: impacts #16 too so a[0]++ works not just v++

ldemailly:
- Adding postfix expressions (++ and --)  Fixes #17 
- changed evaluation order to be more human friendly (appears left to right now)
- remove extra () if the expression is:   `ident = expre` e.g `a = 1 + 2` instead of `a = (1 + 2)` before


ldemailly:
This not quite true postfix version won't quite work for say `arr[i]++`

PR #80, opened 2024-07-29
Keeping line comments on same line.

### LINKED ISSUE #79: keep // comment on same line
right now format puts them on a new line which is typically wrong:

```go
  foo(23) // we need to call with 23 because...
  bar(42)
```
becomes
```go
  foo(23) 
  // we need to call with 23 because...
  bar(42)
```

ldemailly:
Fix #79 

- [x]  adding failing test first
- [x] trim trailing whitespaces
- [x] fix


ldemailly:
(review at your leisure post merge or most fyi)

ccoVeille (approved):
LGTM

PR #81, opened 2024-07-30
Handle block comments /* */

### LINKED ISSUE #14: /* */ style comments

ldemailly:
The difficulty is in the -format output to keep the block comment where they were (same line or new line etc)

This was harder to get right than... I wanted

fixes #14


lexer/lexer.go
```diff
@@ -168,6 +171,33 @@ func (l *Lexer) readLineComment() string {
...
+	var pos2 int
+	if ch == 0 {
+		l.pos--
+		pos2 = l.pos
+	} else {
+		l.pos++
+		pos2 = l.pos
+		/*
+			if l.peekChar() == '\n' { // Add trailing newline to block comment so pretty print with no newline following block comment works.
+				pos2 = l.pos + 1
```
ldemailly, line 194: mystery is here...

ccoVeille (approved):
The code is indeed not fun.

At least, you seem to have added enough test to handle it

eval/eval_test.go
```diff
@@ -20,7 +20,10 @@ func TestEvalIntegerExpression(t *testing.T) {
 		{"10", 10},
 		{"-5", -5},
 		{"-10", -10},
-		{"5 + 5 + 5 + 5 - 10", 10},
+		{"5 + 5 + 5 + 5 - 10 /* some block comment */", 10},
+		/* These don't work, we need to make comment a identity operator or prune them entirely from the AST. */
+		// {"5 + /* block comment in middle of expression */ 2", 7},
+		// {" - /* inline of prefix */ 5", -5},
```
ldemailly, line 26: @ccoVeille also this ^ ie the block/inline comments currently do break expressions, basically act as a ; or \n and making them not do that requires more surgery (like 2 passes, one for -format/pretty print and one for nuking the comments... which seems not ideal)

PR #83, opened 2024-08-01
Add self() for anonymous function recursion

### LINKED ISSUE #82: allow recursion without needing to know the identifier the function object is assigned to
so pure anon functions, eg if we use `self` :

```go
// good 'ole factorial, anonymous
func(n) {
   if n<=1 {
     return 1
   }
   n*self(n-1)
}(5)
```

candidates so far `self()`, `this()`, `$()`

ldemailly:
fixes #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,16 +29,23 @@ Or get one of the [binary releases](https://github.com/grol-io/grol/releases)
 ## What it does
 Sample:
 ```go
-gorepl -parse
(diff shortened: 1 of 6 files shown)


ldemailly:
Screenshot 2024-08-01 at 10 55 06 AM

ldemailly:
fixing the object function print in #84

Issue #38, opened 2024-07-16
add save/load


ldemailly:
serialize deser the state(s) - with some versioning

ldemailly:
probably use https://github.com/klauspost/compress/tree/master/zstd#zstd
eventually with a dictionary

consider whether to normalize, maybe that does a `grol format`

ps: right now some output of printing the ast doesn't actually produce code that can be reprised to identity

will file a new ticket about that part

ldemailly:
`grol -format` is done

also now (#84) there is a compact version - though without comments

probably can use token+len of literal+literal -or- use space or something url save as separator

ldemailly:
Added a "share" button as part of #86 for the wasm/online version

PR #84, opened 2024-08-01
new -compact mode, also use for eval when printing function objects

ldemailly:
```
$ ./grol -format -compact examples/sample.gr  2> /dev/null; echo
unless=macro(cond,iffalse,iftrue){quote(if !unquote(cond){unquote(iffalse)}else{unquote(iftrue)})}unless(10>5,print("BUG: not greater\n"),print("macro test: greater\n"))fact=func(n){log("called fact ",n)if n<=1{return 1}n*self(n-1)}a=[fact(5),"abc",76-3]m={"key":a,73:29}print("m is:",m,"\n")print("Outputting a smiley: 😀\n")first((m["key"]))
```

```
$ ./grol -format -compact examples/sample.gr | ./grol -format -
12:50:10.965 grol dev  go1.22.5 arm64 darwin - welcome!
12:50:10.965 grol dev  go1.22.5 arm64 darwin - welcome!
12:50:10.966 [INF] Formatting stdin
12:50:10.966 [INF] Formatting examples/sample.gr
12:50:10.966 [INF] All done
unless = macro(cond, iffalse, iftrue) {
	quote(if !unquote(cond) {
		unquote(iffalse)
	} else {
		unquote(iftrue)
	})
}
unless(10 > 5, print("BUG: not greater\n"), print("macro test: greater\n"))
fact = func(n) {
	log("called fact ", n)
	if n <= 1 {
		return 1
	}
	n * self(n - 1)
}
a = [fact(5), "abc", 76 - 3]
m = {"key":a, 73:29}
print("m is:", m, "\n")
print("Outputting a smiley: 😀\n")
first((m["key"]))
12:50:10.967 [INF] All done
```

relates a bit to #38

### DIFF
--- a/check_samples_double_format.sh
+++ b/check_samples_double_format.sh
@@ -8,5 +8,10 @@ for file in "$@"; do
     ./grol $file > /tmp/output1
     ./grol /tmp/format2 > /tmp/output2
     diff -u /tmp/output1 /tmp/output2
+    ./grol -format -compact $file > /tmp/format3
+    ./grol -format -compact /tmp/format3 > /tmp/format4
+    diff -u /tmp/format3 /tmp/format4
+    ./grol /tmp/format4 > /tmp/output3
+    diff -u /tmp/output1 /tmp/output3
     echo "---done---"
 done
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ See [Open Issues](https://grol.io/grol/issues) for what's left to do
 ### CLI Usage
 
 ```
-grol 0.25.0 usage:
+grol 0.29.0 usage:
 	grol [flags] *.gr files to interpret or `-` for stdin without prompt
   or no arguments for stdin repl...
 or 1 of the special arguments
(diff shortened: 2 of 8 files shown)


check_samples_double_format.sh
```diff
@@ -8,5 +8,10 @@ for file in "$@"; do
     ./grol $file > /tmp/output1
     ./grol /tmp/format2 > /tmp/output2
     diff -u /tmp/output1 /tmp/output2
+    ./grol -format -compact $file > /tmp/format3
+    ./grol -format -compact /tmp/format3 > /tmp/format4
+    diff -u /tmp/format3 /tmp/format4
+    ./grol /tmp/format4 > /tmp/output3
+    diff -u /tmp/output1 /tmp/output3
```
ldemailly, line 15: this is the test, makes sure the compact form is stable and that it evaluates to same as non compact

probably could have a few unit test too

ldemailly: added unit test, not 100% sure there isn't a corner case but ... hopefully bug would get reported

PR #91, opened 2024-08-02
Allow else if

### LINKED ISSUE #87: add `elseif` (or `else if`)
let's go for `else if` so we don't introduce new keywords

ldemailly:
- [x] else if in parsing
- [x] pretty print them specially

```go
if (i>=3) {10} else if (i>=2) {20} else {30}
```

initially parsed and printed as
```go
if i >= 3 {
	10
} else {
	if i >= 2 {
		20
	} else {
		30
	}
}
```
which was correct but ugly, now
```go
if i >= 3 {
	10
} else if i >= 2 {
	20
} else {
	30
}
```
🎉 

fixes #87

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -303,6 +303,23 @@ type IfExpression struct {
 	Alternative *Statements
 }
 
+func (ie IfExpression) handlElse(out *PrintState) {
+	if out.Compact {
+		out.Print("else")
+	} else {
+		out.Print(" else ")
+	}
+	if len(ie.Alternative.Statements) == 1 && ie.Alternative.Statements[0].Value().Type() == token.IF {
+		// else if
+		if out.Compact {
+			out.Print(" ")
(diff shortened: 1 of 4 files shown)


ldemailly:
this one is pretty cool to look at, it's only a few lines in parser, the rest is test and pretty print handling
(fyi async review @ccoVeille )

ast/ast.go
```diff
@@ -303,21 +303,32 @@ type IfExpression struct {
 	Alternative *Statements
 }
 
+func (ie IfExpression) handlElse(out *PrintState) {
```
ccoVeille, line 306:
```suggestion
func (ie IfExpression) handleElse(out *PrintState) {
```

ldemailly: opsa, thx

ldemailly: #92

PR #92, opened 2024-08-02
implement `func name(...) {....}` as short for `name=func(...) {...}`

### LINKED ISSUE #44: support `func name(...` as alternative to `name = func(...`
would be more typical than all anonymous first class functions

ldemailly:
- fixes #44 
- also follow up on  if/else PR - typo/bad name `handlElse` -> `printElse` / corrects #91

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -204,14 +204,19 @@ func (f Function) Type() Type { return FUNC }
 func (f *Function) SetCacheKey() string {
 	out := strings.Builder{}
 	out.WriteString("func")
+	f.printFunc(&out)
+	f.CacheKey = out.String()
+	return f.CacheKey
+}
+
+// Common part of Inspect and SetCacheKey. Prints the rest of the function.
+func (f *Function) printFunc(out *strings.Builder) {
 	out.WriteString("(")
-	ps := &ast.PrintState{Out: &out, Compact: true}
+	ps := &ast.PrintState{Out: out, Compact: true}
 	ps.ComaList(f.Parameters)
 	out.WriteString("){")
 	f.Body.PrettyPrint(ps)
 	out.WriteString("}")
-	f.CacheKey = out.String()
-	return f.CacheKey
 }
 
 func (f Function) Inspect() string {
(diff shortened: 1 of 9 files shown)


parser/parser.go
```diff
@@ -485,7 +485,13 @@ func (p *Parser) parseBlockStatement() *ast.Statements {
 func (p *Parser) parseFunctionLiteral() ast.Node {
 	lit := &ast.FunctionLiteral{}
 	lit.Token = p.curToken
-
+	// Optional name/identifier
+	if p.peekTokenIs(token.IDENT) {
+		p.nextToken()
+		name := &ast.Identifier{}
+		name.Token = p.curToken
+		lit.Name = name
```
ccoVeille, line 493: It's a bit difficult to read

Maybe this

```suggestion
		lit.Name = &ast.Identifier{Token: p.curToken}
```

ccoVeille: But the rest of the code might be aligned with current code.

object/object.go
```diff
@@ -198,24 +199,34 @@ func WriteStrings(out *strings.Builder, list []Object, before, sep, after string
...
+// so it's not part of the cache key.
 func (f *Function) SetCacheKey() string {
 	out := strings.Builder{}
 	out.WriteString("func")
+	f.CacheKey = f.finishFuncOutput(&out)
+	return f.CacheKey
+}
+
+// Common part of Inspect and SetCacheKey. Outputs the rest of the function.
+func (f *Function) finishFuncOutput(out *strings.Builder) string {
```
ccoVeille, line 212: A bit out of this PR, but maybe you could use godoc link

```suggestion
// Common part of [Inspect] and [SetCacheKey]. Prints the rest of the function.
```

ccoVeille: I'm unsure about the godoc link in such case

ccoVeille (approved):
A few side remarks,otherwise 👍

parser/parser.go
ldemailly: it's like that because the Tokens are in Base... so you'd need to do the super annoying {Base: {Token: ...

object/object.go
ldemailly: godoc doesn't generate doc for private function does it? or maybe there is an option to do so I suppose

parser/parser.go
ccoVeille: Oh indeed

object/object.go
ccoVeille: Good point

PR #93, opened 2024-08-02
Add unique single printable character to Token for serialization

ldemailly:
Toward #38


ldemailly:
Will likely close this for now, put the STRING in wrong section fix in #100 and a few more tweaks like compact mode for DebugString

Issue #41, opened 2024-07-17
add math functions


ldemailly:
interestingly though... sin,cos,abs etc are all easy but... 

what about log? should be ln() I guess to not conflict with log() (logger) or revisit log() or intro prefixes

like math.log

ldemailly:
done in #96

PR #96, opened 2024-08-03
Added support for extensions, including usage and args check - use it to add math functions

### LINKED ISSUE #9: Add "extension" internal functions (calling into a go function), with variadic params, param types etc
right now it's clunky to add token, prefix, and builtin switch

also no support for variadic or types mapping

ldemailly:
example with math.Pow() exposed as pow(float1 float2) (and auto convert from integer)

Screenshot 2024-08-03 at 8 10 55 AM

Fixes #9, #41

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -25,7 +25,7 @@ type State struct {
 
 func NewState() *State {
 	return &State{
-		env:        object.NewEnvironment(),
+		env:        object.NewRootEnvironment(),
 		Out:        os.Stdout,
 		LogOut:     os.Stdout,
 		cache:      NewCache(),
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -59,7 +59,7 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	// let free assignments.
 	id, ok := node.Left.(*ast.Identifier)
 	if !ok {
-		return object.Error{Value: "<assignment to non identifier: " + node.Left.Value().DebugString() + ">"}
+		return object.Error{Value: "assignment to non identifier: " + node.Left.Value().DebugString()}
 	}
 	if rt := right.Type(); rt == object.ERROR {
 		log.Warnf("can't assign %q: %v", right.Inspect(), right)
@@ -88,7 +88,7 @@ func (s *State) evalPostfixExpression(node *ast.PostfixExpression) object.Object
 	id := node.Prev.Literal()
 	val, ok := s.env.Get(id)
 	if !ok {
-		return object.Error{Value: "<identifier not found: " + id + ">"}
+		return object.Error{Value: "identifier not found: " + id}
 	}
 	var toAdd int64
 	switch node.Type() { //nolint:exhaustive // we have default.
@@ -365,11 +365,11 @@ func evalArrayIndexExpression(array, index object.Object) object.Object {
 func (s *State) applyExtension(fn object.Extension, args []object.Object) object.Object {
 	l := len(args)
 	if l < fn.MinArgs {
-		return object.Error{Value: fmt.Sprintf("<wrong number of arguments got=%d, want %s>",
+		return object.Error{Value: fmt.Sprintf("wrong number of arguments got=%d, want %s",
 			l, fn.Inspect())} // shows usage
 	}
 	if fn.MaxArgs != -1 && l > fn.MaxArgs {
-		return object.Error{Value: fmt.Sprintf("<wrong number of arguments got=%d, want %s>",
+		return object.Error{Value: fmt.Sprintf("wrong number of arguments got=%d, want %s",
 			l, fn.Inspect())} // shows usage
 	}
 	for i, arg := range args {
@@ -382,7 +382,7 @@ func (s *State) applyExtension(fn object.Extension, args []object.Object) object
 			continue
 		}
 		if fn.ArgTypes[i] != arg.Type() {
-			return object.Error{Value: fmt.Sprintf("<wrong type of argument got=%s, want %s>",
+			return object.Error{Value: fmt.Sprintf("wrong type of argument got=%s, want %s",
 				arg.Type(), fn.Inspect())}
 		}
 	}
@@ -392,7 +392,7 @@ func (s *State) applyExtension(fn object.Extension, args []object.Object) object
 func (s *State) applyFunction(name string, fn object.Object, args []object.Object) object.Object {
 	function, ok := fn.(object.Function)
 	if !ok {
-		return object.Error{Value: "<not a function: " + fn.Type().String() + ":" + fn.Inspect() + ">"}
+		return object.Error{Value: "not a function: " + fn.Type().String() + ":" + fn.Inspect()}
 	}
 	if v, output, ok := s.cache.Get(function.CacheKey, args); ok {
 		log.Debugf("Cache hit for %s %v", function.CacheKey, args)
@@ -423,7 +423,7 @@ func extendFunctionEnv(name string, fn object.Function, args []object.Object) (*
 	env := object.NewEnclosedEnvironment(fn.Env)
 	n := len(fn.Parameters)
 	if len(args) != n {
-		return nil, &object.Error{Value: fmt.Sprintf("<wrong number of arguments for %s. got=%d, want=%d>",
+		return nil, &object.Error{Value: fmt.Sprintf("wrong number of arguments for %s. got=%d, want=%d",
 			name, len(args), n)}
 	}
 	for paramIdx, param := range fn.Parameters {
@@ -456,7 +456,7 @@ func (s *State) evalIdentifier(node *ast.Identifier) object.Object {
 	}
 	val, ok = s.extensions[node.Literal()]
 	if !ok {
-		return object.Error{Value: "<identifier not found: " + node.Literal() + ">"}
+		return object.Error{Value: "identifier not found: " + node.Literal()}
 	}
 	return val
 }
@@ -471,7 +471,7 @@ func (s *State) evalIfExpression(ie *ast.IfExpression) object.Object {
 		log.LogVf("if %s is object.FALSE, picking else branch", ie.Condition.Value().DebugString())
 		return s.evalInternal(ie.Alternative)
 	default:
-		return object.Error{Value: "<condition is not a boolean: " + condition.Inspect() + ">"}
+		return object.Error{Value: "condition is not a boolean: " + condition.Inspect()}
 	}
 }
 
@@ -518,15 +518,15 @@ func (s *State) evalBangOperatorExpression(right object.Object) object.Object {
 	case object.FALSE:
 		return object.TRUE
 	case object.NULL:
-		return object.Error{Value: "<not of object.NULL>"}
+		return object.Error{Value: "not of object.NULL"}
 	default:
-		return object.Error{Value: "<not of " + right.Inspect() + ">"}
+		return object.Error{Value: "not of " + right.Inspect()}
 	}
 }
 
 func (s *State) evalMinusPrefixOperatorExpression(right object.Object) object.Object {
 	if right.Type() != object.INTEGER {
-		return object.Error{Value: "<minus of " + right.Inspect() + ">"}
+		return object.Error{Value: "minus of " + right.Inspect()}
 	}
 
 	value := right.(object.Integer).Value
@@ -553,7 +553,7 @@ func (s *State) evalInfixExpression(operator token.Type, left, right object.Obje
 	case operator == token.NOTEQ:
 		return object.NativeBoolToBooleanObject(left != right)
 	default:
-		return object.Error{Value: "<operation on non integers left=" + left.Inspect() + " right=" + right.Inspect() + ">"}
+		return object.Error{Value: "operation on non integers left=" + left.Inspect() + " right=" + right.Inspect()}
 	}
 }
 
@@ -568,7 +568,7 @@ func evalStringInfixExpression(operator token.Type, left, right object.Object) o
 	case token.PLUS:
 		return object.String{Value: leftVal + rightVal}
 	default:
-		return object.Error{Value: fmt.Sprintf("<unknown operator: %s %s %s>",
+		return object.Error{Value: fmt.Sprintf("unknown operator: %s %s %s",
 			left.Type(), operator, right.Type())}
 	}
 }
@@ -597,7 +597,7 @@ func evalArrayInfixExpression(operator token.Type, left, right object.Object) ob
 		}
 		return object.Array{Elements: append(leftVal, right.(object.Array).Elements...)}
 	default:
-		return object.Error{Value: fmt.Sprintf("<unknown operator: %s %s %s>",
+		return object.Error{Value: fmt.Sprintf("unknown operator: %s %s %s",
 			left.Type(), operator, right.Type())}
 	}
 }
@@ -623,7 +623,7 @@ func evalMapInfixExpression(operator token.Type, left, right object.Object) obje
 		}
 		return res
 	default:
-		return object.Error{Value: fmt.Sprintf("<unknown operator: %s %s %s>",
+		return object.Error{Value: fmt.Sprintf("unknown operator: %s %s %s",
 			left.Type(), operator, right.Type())}
 	}
 }
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -15,15 +15,22 @@ import (
 )
 
 type State struct {
-	env    *object.Environment
-	Out    io.Writer
-	LogOut io.Writer
-	NoLog  bool // turn log() into println() (for EvalString)
-	cache  Cache
+	env        *object.Environment
+	Out        io.Writer
+	LogOut     io.Writer
+	NoLog      bool // turn log() into println() (for EvalString)
+	cache      Cache
+	extensions map[string]object.Extension
 }
 
 func NewState() *State {
-	return &State{env: object.NewEnvironment(), Out: os.Stdout, LogOut: os.Stdout, cache: NewCache()}
+	return &State{
+		env:        object.NewEnvironment(),
+		Out:        os.Stdout,
+		LogOut:     os.Stdout,
+		cache:      NewCache(),
+		extensions: object.Commands(),
+	}
 }
 
 func (s *State) ResetCache() {
@@ -104,7 +111,7 @@ func (s *State) evalPostfixExpression(node *ast.PostfixExpression) object.Object
 }
 
 // Doesn't unwrap return - return bubbles up.
-func (s *State) evalInternal(node any) object.Object {
+func (s *State) evalInternal(node any) object.Object { //nolint:funlen // we have a lot of cases.
 	switch node := node.(type) {
 	// Statements
 	case *ast.Statements:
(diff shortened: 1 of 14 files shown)


eval/eval.go
```diff
@@ -412,7 +448,12 @@ func (s *State) evalExpressions(exps []ast.Node) ([]object.Object, *object.Error
...
 
 func (s *State) evalIdentifier(node *ast.Identifier) object.Object {
+	// local var can shadow extensions.
 	val, ok := s.env.Get(node.Literal())
+	if ok {
+		return val
+	}
+	val, ok = s.extensions[node.Literal()]
 	if !ok {
 		return object.Error{Value: "<identifier not found: " + node.Literal() + ">"}
```
ccoVeille, line 458: Not related to the current PR, but an sprintf would be easier to read than concatenation

eval/eval.go
```diff
@@ -352,6 +362,32 @@ func evalArrayIndexExpression(array, index object.Object) object.Object {
...
+	for i, arg := range args {
+		if i >= len(fn.ArgTypes) {
+			break
+		}
+		if fn.ArgTypes[i] == object.FLOAT && arg.Type() == object.INTEGER {
+			args[i] = object.Float{Value: float64(arg.(object.Integer).Value)}
+			continue
+		}
+		if fn.ArgTypes[i] != arg.Type() {
+			return object.Error{Value: fmt.Sprintf("<wrong type of argument got=%s, want %s>",
```
ccoVeille, line 384: Not related to this PR only, but do you have error without the `< >` surrounding?

I would add them via `object.Error.Error()` only

eval/eval.go
```diff
@@ -352,6 +362,32 @@ func evalArrayIndexExpression(array, index object.Object) object.Object {
...
+	if fn.MaxArgs != -1 && l > fn.MaxArgs {
+		return object.Error{Value: fmt.Sprintf("<wrong number of arguments got=%d, want %s>",
+			l, fn.Inspect())} // shows usage
+	}
+	for i, arg := range args {
+		if i >= len(fn.ArgTypes) {
+			break
+		}
+		if fn.ArgTypes[i] == object.FLOAT && arg.Type() == object.INTEGER {
+			args[i] = object.Float{Value: float64(arg.(object.Integer).Value)}
```
ccoVeille, line 380: Please add a comment explaining why you are converting to float64 here

eval/eval_test.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"grol.io/grol/ast"
 	"grol.io/grol/eval"
+	_ "grol.io/grol/extensions"
```
ccoVeille, line 8: I'm not fan of this. It uses init and blank import.

https://github.com/leighmcculloch/gochecknoinits

https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html

https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#blank-imports


I would expect the code in this file to call a register method in extensions

extensions/extension.go
```diff
@@ -0,0 +1,34 @@
...
+	"math"
+
+	"grol.io/grol/object"
+)
+
+func init() {
+	Init()
+}
+
+func Init() {
```
ccoVeille, line 13: Do you really need an Init, it could have been in init

Please add a comment if needed

extensions/extension.go
```diff
@@ -0,0 +1,34 @@
...
+	cmd := object.Extension{
+		Name:     "pow",
+		MinArgs:  2,
+		MaxArgs:  2,
+		ArgTypes: []object.Type{object.FLOAT, object.FLOAT},
+		Callback: pow,
+	}
+	err := object.CreateCommand(cmd)
+	if err != nil {
+		panic(err)
```
ccoVeille, line 23: I would return an error (by updating Init signature), an move the panic in init

extensions/extension.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func pow(args []object.Object) object.Object {
```
ccoVeille, line 56: Can't use a variadic or simply have 2 parameter, it might be a huge refactoring but it would be better I think

extensions/extension.go
ccoVeille: Unless you do this

https://github.com/grol-io/grol/pull/96#discussion_r1702804357

extensions/extension.go
```diff
@@ -0,0 +1,34 @@
+package extensions
+
+import (
+	"math"
+
+	"grol.io/grol/object"
+)
+
+func init() {
```
ccoVeille, line 9: I would remove the init that relies on magical code/hack

See
https://github.com/grol-io/grol/pull/96#discussion_r1702804068

object/object.go
```diff
@@ -342,3 +344,39 @@ func (m Macro) Inspect() string {
...
+			coma = ", "
+		}
+		t := strings.ToLower(e.ArgTypes[i-1].String())
+		out.WriteString(fmt.Sprintf("%s%s%d", coma, t, i))
+	}
+	if e.MaxArgs < 0 {
+		out.WriteString(", ...")
+	} else if e.MaxArgs > e.MinArgs {
+		out.WriteString(fmt.Sprintf(", arg%d...arg%d", e.MinArgs+1, e.MaxArgs))
+	}
```
ccoVeille, line 371: You know I hate else, I cannot help myself reporting them

```suggestion
	if e.MaxArgs < 0 {
		out.WriteString(", ...")
		return
	}
	if e.MaxArgs > e.MinArgs {
		out.WriteString(fmt.Sprintf(", arg%d...arg%d", e.MinArgs+1, e.MaxArgs))
	}
```

object/object.go
```diff
@@ -342,3 +344,39 @@ func (m Macro) Inspect() string {
...
+
+func (e *Extension) Usage(out *strings.Builder) {
+	for i := 1; i <= e.MinArgs; i++ {
+		coma := ""
+		if i > 1 {
+			coma = ", "
+		}
+		t := strings.ToLower(e.ArgTypes[i-1].String())
+		out.WriteString(fmt.Sprintf("%s%s%d", coma, t, i))
+	}
```
ccoVeille, line 366: Can't all this be replaced by a strings.Join(whatever, ", ")

I mean the lowercase can be applied on everything no?

ccoVeille (changes requested):

extensions/extension.go
```diff
@@ -0,0 +1,76 @@
...
+}
+func cos(args []object.Object) object.Object {
+	return object.Float{Value: math.Cos(args[0].(object.Float).Value)}
+}
+func tan(args []object.Object) object.Object {
+	return object.Float{Value: math.Tan(args[0].(object.Float).Value)}
+}
+func ln(args []object.Object) object.Object {
+	return object.Float{Value: math.Log(args[0].(object.Float).Value)}
+}
```
ccoVeille, line 76: Maybe you could add an helper such as

```go
func oneArg(f func(float64) float64, o object.Object) object.Float {
           return object.Float{Value: f(o.(object.Float).Value)}
}


func cos(args []object.Object) object.Object { 
         return oneArg(math.Cos, args[0])
 } 
 func tan(args []object.Object) object.Object { 
         return oneArg(math.Tan, args[0])
}
func ln(args []object.Object) object.Object { 
         return oneArg(math.Log, args[0])
 }
```

eval/eval.go
ldemailly: done, thx

eval/eval_test.go
ldemailly: yeah the init() is an issue... not sure how to do extensions some other way though, well I guess I could call extensions.Init() yes

will do that

extensions/extension.go
ldemailly: I usually have Init() callable by unit test (see token.go for a good example) but yes

extensions/extension.go
ldemailly: for one line and a small unwrap/rewrap not sure (but maybe will think about it), on the flip side the exposing them in Init does seem quite repetitive and not happy about that

ccoVeille:
In one commit you said "gofump, instead of gofumpt"

extensions/extension.go
ldemailly: what do you mean? MaxArgs -1 is variadic (and variadic is same as [] anyway)

object/object.go
ldemailly: type can be different

though maybe float1, int2, string3 is weirder, I sort of copied that code from fortio/cli arg help

eval/eval_test.go
ldemailly: done in https://github.com/grol-io/grol/pull/96/commits/2fc25bc43e98d1d01ae13225bc8a61396922a857 ptal

extensions/extension.go
ldemailly: done in https://github.com/grol-io/grol/pull/96/commits/2fc25bc43e98d1d01ae13225bc8a61396922a857 ptal

extensions/extension.go
ldemailly: done

object/object.go
ldemailly: Funny to find someone who hates `else` even more than me :)
Replacing with
```go
	switch {
	case e.MaxArgs < 0:
		out.WriteString(", ...")
	case e.MaxArgs > e.MinArgs:
		out.WriteString(fmt.Sprintf(", arg%d...arg%d", e.MinArgs+1, e.MaxArgs))
	}
```

ldemailly: like it?

extensions/extension.go
ldemailly: table in new version is much much nicer, thanks for pushing

eval/eval.go
ldemailly: So, object.Error aren't quite go errors, they get printed through:

https://github.com/grol-io/grol/blob/v0.33.1/object/object.go#L171
```go
func (e Error) Inspect() string { return "<err: " + e.Value + ">" }
```
which does add <err ...> around, so I can drop some of the extra in the text itself indeed

ldemailly: done

eval/eval.go
ldemailly: well not for 2 strings imo (removed the `+ ">"` now so should be better)

extensions/extension.go
ccoVeille: It could be pow(val, exp)

Or pow(args ...object.Object)

extensions/extension.go
```diff
@@ -71,25 +39,28 @@ func initInternal() error {
...
+		{math.Sin, "sin"},
+		{math.Cos, "cos"},
+		{math.Tan, "tan"},
+		{math.Log, "ln"}, // proper name for natural logarithm and also doesn't conflict with logger builtin.
+		{math.Sqrt, "sqrt"},
+		{math.Exp, "exp"},
+		{math.Asin, "asin"},
+		{math.Acos, "acos"},
+		{math.Atan, "atan"},
+		{math.Round, "round"},
```
ccoVeille, line 55: I would add floor and ceil

object/object.go
ccoVeille: Ok

extensions/extension.go
ldemailly: yes... so that would be like it is today MinArgs=MaxArgs=2

the second one is MinArgs=0 MaxArgs=-1

extensions/extension.go
```diff
@@ -0,0 +1,79 @@
+package extensions
+
+import (
+	"math"
+
+	"grol.io/grol/object"
+)
+
+var (
+	initDone  = false
```
ccoVeille, line 10: Maybe an atomic.Bool?

Or sync.Once in init?

object/object.go
ldemailly: maybe I should reset `i` each time type changes but then if it's

int, float, int -> int1, float1, int2

is a bit harder... maybe I drop the `i` altogether

object/object.go
ccoVeille: Yes

object/object.go
ccoVeille: It's worth trying

object/interp.go
```diff
@@ -0,0 +1,36 @@
...
+func Init() {
+	commands = make(map[string]Extension)
+}
+
+func CreateCommand(cmd Extension) error {
+	if cmd.Name == "" {
+		return errors.New("empty command name")
+	}
+	if cmd.MaxArgs != -1 && cmd.MinArgs > cmd.MaxArgs {
+		return errors.New(cmd.Name + ": min args > max args")
```
ccoVeille, line 22: Come on 😅


```suggestion
		return fmt.Errorf("%s: min args > max args", cmd.Name)
```

(Here and in a few other place)

object/type_string.go
```diff
@@ -21,12 +21,13 @@ func _() {
...
-	_ = x[LAST-13]
+	_ = x[EXTENSION-13]
+	_ = x[LAST-14]
 }
 
-const _Type_name = "UNKNOWNINTEGERFLOATBOOLEANNILERRORRETURNFUNCSTRINGARRAYMAPQUOTEMACROLAST"
+const _Type_name = "UNKNOWNINTEGERFLOATBOOLEANNILERRORRETURNFUNCSTRINGARRAYMAPQUOTEMACROEXTENSIONLAST"
 
-var _Type_index = [...]uint8{0, 7, 14, 19, 26, 29, 34, 40, 44, 50, 55, 58, 63, 68, 72}
+var _Type_index = [...]uint8{0, 7, 14, 19, 26, 29, 34, 40, 44, 50, 55, 58, 63, 68, 77, 81}
```
ccoVeille, line 30: This is cryptic

Can't you use something else?

I mean relying on a huge string and magical offset?

Please at least  add a comment about such "hack"

wasm/wasm_main.go
```diff
@@ -54,5 +55,9 @@ func main() {
 	global := js.Global()
 	global.Set("grol", js.FuncOf(jsEval))
 	global.Set("grolVersion", js.ValueOf(grolVersion))
+	err := extensions.Init()
+	if err != nil {
+		log.Critf("Error initializing extensions: %v", err)
```
ccoVeille, line 60: First time seeing Crit in your code, how does it differ from Fatal?

I mean you are using Fatal in main.go when using extension.Init

object/interp.go
```diff
@@ -0,0 +1,36 @@
...
+
+var commands map[string]Extension
+
+func init() {
+	Init()
+}
+
+func Init() {
+	commands = make(map[string]Extension)
+}
```
ccoVeille, line 15: The init hack should be removed here too

extensions/extension.go
ldemailly: none of the code is thread safe so far and even if it was, Init() style should be called from the main thread (like init() is)

object/type_string.go
ldemailly: read the "generated code, do not edit" at the top:)

this is done by go:generate and stringer, unfortunately it's checked in because go can't agree to add some go:generate at build/install time support

wasm/wasm_main.go
ldemailly: yeah I changed it to crit because in wasm it complains there is no os.exit (was fatalf earlier in the pr before I tested)

object/interp.go
ccoVeille: perfprint linter?

I tend to use sprintf way for readability purpose, I hated this pattern a few years ago. Not it seems natural, especially as it helps with i18n. But it's not the point here

object/type_string.go
ccoVeille: OK … so cryptic …

object/interp.go
ldemailly: if only maps had non useless nil zero values, it's annoying but let me see (again see token.go where init is mandatory)

wasm/wasm_main.go
ccoVeille: Why not changing it in main.go then?

object/interp.go
ccoVeille: OK

extensions/extension.go
ccoVeille: OK, maybe a comment about it then

object/type_string.go
ldemailly: the code it generates is pretty cool... just that... I wish I didn't have to checkin generated code, I hate doing that

wasm/wasm_main.go
ldemailly: because in main.go I want it to die if there is a bug with extensions

object/interp.go
ldemailly: I could lazy init in first CreateCommand() I guess but... not sure that's better, it'll be more code

wasm/wasm_main.go
ccoVeille: But then won't wasm complain about it?

I might have missed something

object/interp.go
```diff
@@ -0,0 +1,35 @@
...
+var commands map[string]Extension
+
+func Init() {
+	commands = make(map[string]Extension)
+}
+
+func CreateCommand(cmd Extension) error {
+	if commands == nil {
+		Init()
+	}
```
ccoVeille, line 16: What about this?

I mean fonuou need an Init?

```suggestion
func CreateCommand(cmd Extension) error {
	if commands == nil {
		  commands = make(map[string]Extension)
	}
```

object/interp.go
ldemailly: fixed it

object/interp.go
ldemailly: I keep Init() so I and others can reset the list of functions (I haven't done it yet in object_test but I do in token_test)

ldemailly: added godoc and better names in latest commit

wasm/wasm_main.go
ldemailly: `grol/wasm/wasm_main.go` is the program (main) built for wasm, it's different from `grol/main.go` (because it creates JS apis etc)

object/interp.go
ldemailly: > perfprint linter?

yeah probably, and I did change some in fortio when that linter first showed up

I added the cmd.Name because when it was panic() you could tell which command was wrong, if any, but now that it just bubbles up as error, the cmd.Name is needed - if I had started from fmt.Errorf (which I kinda did and then a linter told me to switch to `errors.New`...) I would agree but, it'll stay like this for now :)

object/object.go
ldemailly: done

extensions/extension.go
ldemailly: did add some good

extensions/extension.go
ldemailly: See https://github.com/grol-io/grol/pull/96/commits/7d759b8137dc5947bb372e3de1dd079ba5de80e3 for `sprintf` integration as variadic example

extensions/extension.go
ldemailly: I think round and trunc are enough but sure, adding

ldemailly: added

ldemailly:
@ccoVeille I'm working on a grol side variadic support off this branch but... I think we should merge this first?

ccoVeille (approved):

Issue #56, opened 2024-07-21
tour of grol


ldemailly:
using the wasm for interactive would be nice

probably need a more widgety version of the fragment used in grol.io main page

ldemailly:
consider using

https://codemirror.net/ 
or
https://microsoft.github.io/monaco-editor/

ldemailly:
https://blog.replit.com/code-editors

ldemailly:
having added `info` helps for discoverability

ldemailly:
we should group the function list by categories (maybe introduce that category in the extension struct)

PR #99, opened 2024-08-04
Variadic grol side

### LINKED ISSUE #97: add some grol stdlib in grol
for instance
```go
func abs(x) {
	if x < 0 {
		-x
	} else {
		x
	}
}
```

ldemailly:
So `..` is a magic argument that both indicate that a function is variadic and also receives the array of optional arguments within the function

Probably need to document that last argument if an array of objects is expended for both grol and extension variadic functions, this makes:
```go
printf=func(format, ..){print(sprintf(format, ..))}
```
work but `printf(["%d\n", 42])` also works, ie any array not just `..`

Also fixes #97


lexer/lexer_test.go
```diff
@@ -42,6 +42,7 @@ j--
 /*/*/
 /* This is a
    multiline comment */
+..
```
ccoVeille, line 45: Please add something about

1..2
1 .. 3
1 .. 4
1.. 5

To make sure how the lexer reacts with things that could be interpreted as number

And maybe

1.

ccoVeille, line 45: Please add the following to test

...
. ..
.. .
....

… (a Unicode ellipsis)

My idea is to make sure nothing strange could pop out from a future refactoring

lexer/lexer.go
```diff
@@ -72,12 +72,18 @@ func (l *Lexer) NextToken() *token.Token {
 		} else {
 			return token.EOFT
 		}
+	case '.':
+		if nextChar == '.' { // DOTDOT
+			l.pos++
+			return token.ConstantTokenChar2(ch, nextChar)
+		}
+		// number can start with . eg .5
+		return token.Intern(l.readNumber(ch))
```
ccoVeille, line 81: No check on isDigit before ?

See https://github.com/grol-io/grol/pull/99#discussion_r1703056718

lexer/lexer.go
```diff
@@ -72,12 +72,18 @@ func (l *Lexer) NextToken() *token.Token {
...
 	default:
 		switch {
```
ccoVeille, line 83: I would have expected something like that


```suggestion
	case '.':
		if nextChar == '.' { // DOTDOT
			l.pos++
			return token.ConstantTokenChar2(ch, nextChar)
		}
		fallthrough
 	default:
 		switch {
```

or


```suggestion

	default:
		if ch == '.' && nextChar == '.' { // DOTDOT
			l.pos++
			return token.ConstantTokenChar2(ch, nextChar)
		}
 	
 		switch {
```

ccoVeille: Please add tests according to this

README.md
```diff
@@ -64,6 +64,10 @@ macros and more all the time (like canonical reformat using `grol -format` and w
 
 automatic memoization
 
+easy extensions/adding go functions to grol (see [extensions/extension.go](extensions/extension.go) for a lot of `math` additions)
```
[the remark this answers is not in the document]
ldemailly: adding but given these sentences fragment are somewhat runaway and don't have Uppercase it'll look odd

eval/eval.go
```diff
@@ -712,3 +747,20 @@ func evalFloatInfixExpression(operator token.Type, left, right object.Object) ob
 		return object.Error{Value: "unknown operator: " + operator.String()}
 	}
 }
+
+// Adds a (grol) from go to the base identifiers.
```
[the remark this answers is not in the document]
ldemailly:
```suggestion
// Adds an evaluated result like a function object, from Go, to the base identifiers.
```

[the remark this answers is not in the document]
ldemailly: indeed, wonder how that happened... fixing

lexer/lexer.go
ldemailly: could have been "passthrough" but I already know it started with . so I don't want to redo the letter check

lexer/lexer.go
ldemailly: there are tests already for .3 etc

lexer/lexer_test.go
ldemailly: there are no range yet (1. is there already in many places)

1..2 is actually
```
1.
.2
```
but it's always been that

1 .. 2

complains about .. being undefined (unless used inside a varargs or set before)

short version is we need #56 and more documentation about the language

ldemailly:
been trying to apply batch of 6 or 7 suggestions for the 3rd time already, getting "can't have more than 1 suggestion per line" bogus GitHub error

eval/eval.go
```diff
@@ -421,14 +437,33 @@ func (s *State) applyFunction(name string, fn object.Object, args []object.Objec
...
 	}
-	for paramIdx, param := range fn.Parameters {
+	for paramIdx, param := range params {
 		env.Set(param.Value().Literal(), args[paramIdx])
 	}
+	if fn.Variadic {
+		env.Set("..", extra)
+	}
 	// for recursion in anonymous functions.
 	// TODO: consider not having to keep setting this in the function's env and treating as a keyword.
```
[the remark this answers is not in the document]
ldemailly: and? not human doesn't mean the function doesn't have an env

eval/eval.go
```diff
@@ -752,7 +752,7 @@ func evalFloatInfixExpression(operator token.Type, left, right object.Object) ob
 	}
 }
 
-// Adds a (grol) from go to the base identifiers.
+// Adds a (grol) from Go to the base identifiers.
```
ldemailly, line 755: stupid UI picked the wrong one to commit

eval/eval.go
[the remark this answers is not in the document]
ldemailly: and github lost my fix !&#%&&%

README.md
```diff
@@ -64,6 +64,10 @@ macros and more all the time (like canonical reformat using `grol -format` and w
...
+
+variadic functions both go side and grol side (using `..` on grol side)
```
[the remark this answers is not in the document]
ldemailly: mostly because of the lexer which currently never needs to look at more than 2 characters at a time, also it's different in how go name's the "..." (though maybe both are bad reason and should use 3 dots anyway)

lexer/lexer_test.go
ldemailly: fuzz testing would be nice too

ldemailly: #101 for fancier fuzz test

lexer/lexer.go
ldemailly: first proposal is worse as it does a check for letter when we already know there is a `.`

and same for 2nd one, also all the other 2 token check are in the main switch, not in default/fallback
the start with digit or number is I think actually clearer than with the before `|| ch = '.'`

and I don't want to have `ch == '.'` checked twice, I prefer to repeat token.Intern(l.readNumber(ch))

lexer/lexer.go
ldemailly: short version is unlike some of the other scripting language that force you to type `0.5`, grol is fine with `.5` (had to handle that in readNumber, including that it doesn't allow 2 dots etc... see that code and its tests)

eval/eval.go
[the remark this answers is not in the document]
ccoVeille: my point is the fact possessive `'s` cannot be used by object.

> The possessive form is used with nouns referring to people, groups of people, countries, and animals.

https://www.ef.com/wwen/english-resources/english-grammar/forming-possessive/#:~:text=To%20form%20the%20possessive%2C%20add,an%20apostrophe%20after%20the%20s.&text=For%20names%20ending%20in%20s,s%2C%20or%20just%20an%20apostrophe.

README.md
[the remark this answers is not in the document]
ccoVeille: I'm fine with it. I just asked you to document it as a comment

lexer/lexer_test.go
ccoVeille: > there are no range yet

Not yet,y point was to add these to tests and see what happens.

So yes,the fuzzy testing might help

lexer/lexer.go
ccoVeille: I might be wrong in the way I reported it

But my point was that code let me think there could be a problem with

1.a
.a
1.-
4. (4 dot space)

The current code seem to me to be:
- we found a dot
- let's read the number that follows

But what comes next might not be a number

Maybe it's something that readNumber handles well. But with previous code the one in the default case, the type was checked

ccoVeille:
What about adding to number tests things that could come later

1.3e6
1+6i

Adding them to test would raise errors, but then you will think about fixing later

lexer/lexer.go
ldemailly: > we found a dot let's read the number that follows

that's what the old code did too, if we agree that new code and old code are exactly the same except when `..` is found we can then separate the issue of what happens with invalid numbers from the structure of the code

old (grol 0.31.0) code:
```
$ .a
11:49:35.063 [CRI] parser has 1 error(s)
11:49:35.063 [ERR] parser error: could not parse "." as float
```
same with new code and it'll be the same for all your examples as they don't involve 2 `..` 
(and again there are lots of test about these, like 4. space is 4.0 float and spaces are ignored, etc)

ldemailly:
you filled a follow up for _ and 1e6 and I filled for fuzz testing, can we merge this? 

it's quite a big addition, and so is #102 and I'd like to get both in (as long as there is no terrible tech debt or bug)

ccoVeille:
Yes, they can be merged

PR #102, opened 2024-08-04
ALL_CAPS are now constants. Changed/shared recursive function environment. Fixed bug with macro environment. Added pprof.

### LINKED ISSUE #47: clarify variable scope, as in...
can read from any scope but only write in current one

### LINKED ISSUE #98: make UPPERCASE constants (you can set once but not change)
That would protect PI and E coming in #96

ldemailly:
fixes #98 

https://github.com/grol-io/grol/issues/47#issuecomment-2267631006

fixes #47

Also fixes mix up with macro environment which was incorrectly reusing the full eval.State when it really only needs object.Environment, split the two. 

This showed up as a bug with ==> macro showing up in output even if no macro was defined (because it got PI and E from other MR)

Added `-profile-mem` and `-profile-cpu` options

Figured out regression and saved as examples/pi_perf.gr

Handle tabs in wasm share button

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -68,7 +68,7 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 		return right
 	}
 	log.LogVf("eval assign %#v to %#v", right, id.Value())
-	return s.env.Set(id.Literal(), right)
+	return s.env.Set(id.Literal(), right) // Propagate possible error (constant setting).
 }
 
 func ArgCheck[T any](msg string, n int, vararg bool, args []T) *object.Error {
@@ -104,7 +104,7 @@ func (s *State) evalPostfixExpression(node *ast.PostfixExpression) object.Object
 	case object.Integer:
 		return s.env.Set(id, object.Integer{Value: val.Value + toAdd})
 	case object.Float:
-		return s.env.Set(id, object.Float{Value: val.Value + float64(toAdd)})
+		return s.env.Set(id, object.Float{Value: val.Value + float64(toAdd)}) // So PI++ fails not silently.
 	default:
 		return object.Error{Value: "can't increment/decrement " + val.Type().String()}
 	}
@@ -174,7 +174,7 @@ func (s *State) evalInternal(node any) object.Object {
 		if name != nil {
 			oerr := s.env.Set(name.Literal(), fn)
 			if oerr.Type() == object.ERROR {
-				return oerr
+				return oerr // propagate that func FOO() { ... } can only be defined once.
 			}
 		}
 		return fn
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -116,7 +116,7 @@ func (s *State) evalPostfixExpression(node *ast.PostfixExpression) object.Object
 }
 
 // Doesn't unwrap return - return bubbles up.
-func (s *State) evalInternal(node any) object.Object {
+func (s *State) evalInternal(node any) object.Object { //nolint:funlen // quite a lot of cases.
 	switch node := node.(type) {
 	// Statements
 	case *ast.Statements:
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -448,9 +448,11 @@ func extendFunctionEnv(
 ) (*object.Environment, *object.Error) {
 	// https://github.com/grol-io/grol/issues/47
 	// fn.Env is "captured state" but for recursion we now parent from current state; eg
-	// func test(n) {if (n==2) {x=1}; if (n==1) {return x}; test(n-1)}; test(3)
+	//     func test(n) {if (n==2) {x=1}; if (n==1) {return x}; test(n-1)}; test(3)
 	// return 1 (state set by recursion with n==2)
-	env := object.NewFunctionEnvironment(fn, currrentEnv)
+	// Make sure `self` is used to recurse, or named function, otherwise the function will
+	// need to be found way up the now much deeper stack.
+	env, sameFunction := object.NewFunctionEnvironment(fn, currrentEnv)
 	params := fn.Parameters
 	atLeast := ""
 	extra := object.Array{}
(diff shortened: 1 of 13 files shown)


eval/eval.go
```diff
@@ -427,7 +428,12 @@ func extendFunctionEnv(name string, fn object.Function, args []object.Object) (*
 			name, len(args), n)}
 	}
 	for paramIdx, param := range fn.Parameters {
-		env.Set(param.Value().Literal(), args[paramIdx])
+		oerr := env.Set(param.Value().Literal(), args[paramIdx])
+		log.LogVf("set %s to %s - %s", param.Value().Literal(), args[paramIdx].Inspect(), oerr.Inspect())
+		if oerr.Type() == object.ERROR { // TODO: doesn't trigger
+			oe, _ := oerr.(object.Error)
+			return nil, &oe
```
ldemailly, line 435: This doesn't seem to trigger...

eval/eval.go
```diff
@@ -435,8 +441,18 @@ func (s *State) applyFunction(name string, fn object.Object, args []object.Objec
...
 
-func extendFunctionEnv(name string, fn object.Function, args []object.Object) (*object.Environment, *object.Error) {
-	env := object.NewEnclosedEnvironment(fn.Env)
+func extendFunctionEnv(
+	currrentEnv *object.Environment,
+	name string, fn object.Function,
+	args []object.Object,
+) (*object.Environment, *object.Error) {
+	// https://github.com/grol-io/grol/issues/47
+	// fn.Env is "captured state" but for recursion we now parent from current state; eg
```
ccoVeille, line 450:
```suggestion
	// fn.Env is "captured state", but for recursion we now parent from current state; eg
```

Is parent used as a verb?

eval/eval_test.go
```diff
@@ -11,6 +11,13 @@ import (
 	"grol.io/grol/parser"
 )
 
+func init() {
+	err := extensions.Init()
+	if err != nil {
+		panic(err)
+	}
+}
```
ccoVeille, line 19: Nope


```suggestion
func TestMain(m *testing.M) {	
	err := extensions.Init()
	if err != nil {
		panic(err)
	}
	os.Exit(m.Run())
}
```

main.go
```diff
@@ -34,9 +39,22 @@ func Main() int {
...
+		if err != nil {
+			log.Fatalf("can't start cpu profile: %v", err)
+		}
+		log.Infof("Writing cpu profile to %s", *cpuprofile)
+		defer pprof.StopCPUProfile()
+	}
 	err := extensions.Init()
 	if err != nil {
-		log.Fatalf("Error initializing extensions: %v", err)
+		return log.FErrf("Error initializing extensions: %v", err)
```
ccoVeille, line 57: So FErrf is Fatalf ?

The name could be better

ccoVeille, line 57: Why using your ferry here, and using Fatalf in previous if block you add

ccoVeille (approved):
I'm happy you have tests, because it is not a fun PR

main.go
ldemailly: ?

eval/eval_test.go
ldemailly: I realize you don't like init() (for good reasons in library or main code) but how is that necessary or different or better for a test package

main.go
ldemailly: ic you mean why not change all of them, only changed what the linter complained about re defer not to running but probably should be all yes, in theory/for consistency

ldemailly: done

eval/eval_test.go
ccoVeille: Here it's not only because init, it's because there is something made to prepend things in tests

main.go
ccoVeille: Thanks

eval/eval_test.go
ldemailly: I did anyway :) thx

main.go
ldemailly: of note there are still a few more log.Fatalf in other functions (like if you fail to pass a valid filename) and it's fine... for now at least (just means some pprof defer won't happen)

eval/eval.go
```diff
@@ -101,19 +100,23 @@ func (s *State) evalPostfixExpression(node *ast.PostfixExpression) object.Object
...
 	}
+	if oerr.Type() == object.ERROR {
+		return oerr
+	}
 	return val
 }
 
 // Doesn't unwrap return - return bubbles up.
-func (s *State) evalInternal(node any) object.Object {
+func (s *State) evalInternal(node any) object.Object { //nolint:funlen // quite a lot of cases.
```
ccoVeille, line 119: Maybe you could add a map[type]func(Node) Object 

Then you create one method per type

So evalInternal method will then simply check if there is a func, then call it

object/state.go
```diff
@@ -28,12 +39,58 @@ func (e *Environment) Get(name string) (Object, bool) {
...
 
-func (e *Environment) Set(name string, val Object) Object {
+// Defines constant as all CAPS (with _ ok in the middle) identifiers.
+// Note that we use []byte as all identifiers are ASCII.
+func Constant(name string) bool {
+	for i, v := range name {
+		if i != 0 && v == '_' {
+			continue
+		}
+		if v < 'A' || v > 'Z' {
```
ccoVeille, line 49: Constant may/could also contains number, no?

ldemailly: well... not right now indeed :) PI2=PI; PI2++ works https://grol.io/?c=PI2=PI+PI2%2B%2BPI2

eval/eval.go
ccoVeille: What do you think about this?

ldemailly: Thanks for the reminder, I had skipped that one earlier

If it worked, we'd need to profile it (good that there is `-profile-cpu` now) but I don't think it'd help because you'd need to cast anyway inside each small function (but try and let me know :) ?)

Note that for other types, like token.Type or object.Type an array lookup would be good, as the type fit into 256 and I had started something like that in #93

PR #107, opened 2024-08-05
added `info` returning info map. made object.Array sortable replacing MapKeys, use Equals in constant check, fix missing FLOAT handling

ldemailly:
- added `info` returning info map including identifiers stack (see README.md update for examples) and:
- made object.Array sortable replacing MapKeys
- use Equals in constant check and change objects Equals to stay native `bool` until last minute
- fix missing FLOAT handling

Example:
```go
$ info
{"all_ids":{0:["E","PI","abs","log2","printf"]},"gofuncs":["acos","asin","atan","ceil","cos","exp","floor","ln","log10","pow","round","sin","sprintf","sqrt","tan","trunc"],"keywords":["else","error","false","first","func","if","len","log","macro","print","println","quote","rest","return","true","unquote"],"platform":"dev  go1.22.5 arm64 darwin","tokens":["!","!=","%","(",")","*","+","++",",","-","--","..","/",":",":=",";","<","<=","=","==",">",">=","[","]","{","}"],"version":"dev"}
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,21 @@ $ m=fact(7)
 $ m/n
 == Parse ==> m / n
 == Eval  ==> 7
-$ info["all_ids"][0]
-["E","PI","abs","log2","printf"]
+$ func fx(n) {if n>0 {return fx(n-1)}; info["all_ids"]}; fx(3)
+== Parse ==> func fx(n) {
+	if n > 0 {
+		return fx(n - 1)
+	}
+	(info["all_ids"])
+}
+fx(3)
+== Eval  ==> {0:["E","PI","abs","fx","log2","printf"],1:["n","self"],2:["fx","n","self"],3:["fx","n","self"],4:["fx","n","self"]}
 $ info["gofuncs"]
-["acos","asin","atan","ceil","cos","exp","floor","ln","log10","pow","round","sin","sprintf","sqrt","tan","trunc"]
+== Parse ==> (info["gofuncs"])
+== Eval  ==> ["acos","asin","atan","ceil","cos","exp","floor","ln","log10","pow","round","sin","sprintf","sqrt","tan","trunc"]
 $ info["keywords"]
-["else","error","false","first","func","if","len","log","macro","print","println","quote","rest","return","true","unquote"]
+== Parse ==> (info["keywords"])
+== Eval  ==> ["else","error","false","first","func","if","len","log","macro","print","println","quote","rest","return","true","unquote"]
 ```
 
 ## Language features
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ easy extensions/adding Go functions to grol (see [extensions/extension.go](exten
 
 variadic functions both Go side and grol side (using `..` on grol side)
 
+Use `info` to see all the available functions, keywords, operators etc... (can be used inside functions too to examine the stack)
+
 See also [sample.gr](examples/sample.gr) and others in that folder, that you can run with
 ```
 gorepl examples/*.gr
@@ -169,4 +171,4 @@ flags:
   -shared-state
     	All files share same interpreter state (default is new state for each)
 ```
-(excluding logger control, see `gorepl help` for all the flags, of note `-logger-no-color` will turn off colors for gorepl too)
+(excluding logger control, see `gorepl help` for all the flags, of note `-logger-no-color` will turn off colors for gorepl too, for development there are also `-profile*` options for pprof)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -68,6 +83,8 @@ easy extensions/adding Go functions to grol (see [extensions/extension.go](exten
 
 variadic functions both Go side and grol side (using `..` on grol side)
 
+Use `info` to see all the available functions, keywords, operators etc... (can be used inside functions too to examine the stack)
```
ccoVeille, line 86: OK would have expected to use "help"

But why not

object/state.go
```diff
@@ -31,7 +36,63 @@ func (e *Environment) Len() int {
...
+		e = e.outer
+	}
+	info := e.BaseInfo()
+	info[String{"all_ids"}] = allKeys
+	// TODO: Cache this as it can't change without recompiling:
+	return info
+}
+
 func (e *Environment) Get(name string) (Object, bool) {
+	if name == "info" {
```
ccoVeille, line 93: Do you plan to be able to use info with argument, such as "info sin"

Or "help sin"

object/state.go
```diff
@@ -63,7 +124,7 @@ func (e *Environment) Set(name string, val Object) Object {
 		old, ok := e.Get(name) // not ok
 		if ok {
 			log.Infof("Attempt to change constant %s from %v to %v", name, old, val)
-			if !Hashable(old) || !Hashable(val) || old != val {
+			if Equals(old, val) == FALSE {
```
ccoVeille, line 127: The uppercase false caught my attention. Shouldn't be this?

```suggestion
			if !Equals(old, val) {
```

README.md
ldemailly: it's my tcl history, but more importantly, it returns a map, that can be accessed, it's not a help

we could build a help or even command completion, based on that map

object/state.go
ldemailly: functions being objects, you don't need "help sin" you can just type "sin":

Screenshot 2024-08-05 at 12 45 58 PM

and to repeat it's a map; eg info["gofuncs"] is just the go function extensions array

object/state.go
ldemailly: Equals() is returning an Object and thus TRUE or FALSE not a boolean directly (there is conversion function)

(because it's used in the `==` operator for instance where objects are needed not go primitive types)

ldemailly: maybe I can do the wrapping elsewhere though/will check

object/state.go
ldemailly: we can add help text later though, that could be useful (keyed of that map)

object/state.go
ldemailly: but you're right, it was ugly, changed to do NativeBoolToBooleanObject() only in leaves/where needed

thanks!

object/object.go
```diff
@@ -67,50 +67,52 @@ func NativeBoolToBooleanObject(input bool) Boolean {
...
 
-func Equals(left, right Object) Object {
+func Equals(left, right Object) bool {
 	if left.Type() != right.Type() {
-		return FALSE
+		return false
 	}
 	switch left := left.(type) {
 	case Integer:
-		return NativeBoolToBooleanObject(left.Value == right.(Integer).Value)
```
ccoVeille, line 76: Are you still using NativeBoolToBooleanObject func?

If not maybe, it can be deleted

ccoVeille: Same question for TRUE and FALSE objects

ldemailly: see the rest of the diff... yes it's needed because inside grol that's the objects for `true` and `false`

eval/eval.go
```diff
@@ -671,9 +671,9 @@ func evalMapInfixExpression(operator token.Type, left, right object.Object) obje
 	rightMap := right.(object.Map)
 	switch operator { //nolint:exhaustive // we have default.
 	case token.EQ:
-		return object.MapEquals(leftMap, rightMap)
+		return object.NativeBoolToBooleanObject(object.MapEquals(leftMap, rightMap))
```
ldemailly, line 674: like here @ccoVeille

object/state.go
ldemailly: #108

eval/eval.go
ccoVeille: I had already marked the file as read, so when I saw the refactoring I thought it was no longer used

ccoVeille (approved):

eval/eval.go
ldemailly: np and btw thanks a lot for all your time reviewing all these PRs

PR #114, opened 2024-08-06
Add json(), unjson(), eval(). `println(json(info))` works despite int keys in the all_ids stack. Revamped errors vs normal output, don't output "nil" if not interactive. Updated extensions API.

### LINKED ISSUE #115: (Some) Extensions need current eval context
So we need to revamp the call back and maybe use

Environment, FuncName, Args

instead of current just Args (that doesn't include function name; which may not be present)

              For eval it highlights we should be able to attach to the current environment (instead of eval in thin air/different new one)

_Originally posted by @ldemailly in https://github.com/grol-io/grol/issues/114#issuecomment-2270506795_

ldemailly:
Add json() though... ugly but print(json(info)) works.

Changed that errors are not put to stdout and nil result aren't printed except in interactive mode

Also fixed #115 by adding a new LongCallBack option that takes environment and fname as additional args. In theory should be the eval state but that creates a module loop...

```json
$ ./grol -quiet -c "println(json(info))" | jq
{
  "all_ids": {
    "0": [
      "E",
      "PI",
      "abs",
      "log2",
      "printf"
    ]
  },
  "gofuncs": [
    "acos",
    "asin",
    "atan",
    "ceil",
    "cos",
    "eval",
    "exp",
    "floor",
    "json",
    "ln",
    "log10",
    "pow",
    "round",
    "sin",
    "sprintf",
    "sqrt",
    "tan",
    "trunc",
    "unjson"
  ],
  "keywords": [
    "else",
    "error",
    "false",
    "first",
    "func",
    "if",
    "len",
    "log",
    "macro",
    "print",
    "println",
    "quote",
    "rest",
    "return",
    "true",
    "unquote"
  ],
  "platform": "dev  go1.22.5 arm64 darwin",
  "tokens": [
    "!",
    "!=",
    "%",
    "(",
    ")",
    "*",
    "+",
    "++",
    ",",
    "-",
    "--",
    "..",
    "/",
    ":",
    ":=",
    ";",
    "<",
    "<=",
    "=",
    "==",
    ">",
    ">=",
    "[",
    "]",
    "{",
    "}"
  ],
  "version": "dev"
}
```


extensions/extension.go
```diff
@@ -115,3 +126,26 @@ func sprintf(args []object.Object) object.Object {
...
+
+func convertMap(m map[any]any) map[string]any {
+	result := make(map[string]any, len(m))
+	for key, value := range m {
+		if valueMap, ok := value.(map[any]any); ok {
+			value = convertMap(valueMap)
+		}
+		result[fmt.Sprint(key)] = value
+	}
+	return result
```
ldemailly, line 138: Sad that this horror is needed

Considering just doing my own serialization instead

ldemailly:
For eval it highlights we should be able to attach to the current environment (instead of eval in thin air/different one)

object/object.go
```diff
@@ -390,9 +402,21 @@ type Extension struct {
...
+
+// ShortExtFunction is the signature for callbacks that do not need more than the arguments (like math functions).
+type ShortExtFunction func(args []Object) Object
+
 // ExtFunction is the signature of what grol will call when the extension is invoked.
 // Incoming arguments are validated for type and number of arguments based on [Extension].
-type ExtFunction func(args []Object) Object
+// eval is the opaque state passed from the interpreter, it can be used with eval.Eval etc
+// name is the function name as registered under.
+type ExtFunction func(eval any, name string, args []Object) Object
```
ldemailly, line 419: I would like to be able to put `eval.State` here but that creates a module loop

extensions/extension.go
```diff
@@ -115,3 +141,42 @@ func sprintf(args []object.Object) object.Object {
...
 	return object.String{Value: res}
 }
+
+func convertMap(m map[any]any) map[string]any {
+	result := make(map[string]any, len(m))
+	for key, value := range m {
+		if valueMap, ok := value.(map[any]any); ok {
+			value = convertMap(valueMap)
+		}
+		result[fmt.Sprint(key)] = value
```
ccoVeille, line 151: Maybe with the type because 1 and 1. and true may collide no?

Like 1, 1., and "1"

Maybe it's OK, but I'm raising the point

ccoVeille (commented):
Uneasy, but LGTM

extensions/extension.go
ldemailly: I'm dropping this code in favor of handcoded serialization for all objects (commit incoming, was just debugging why I still saw some encoding/json in GoSizeAnalyzer and it turned out it was from fortio.org/sets, I just made a new release that has the `no_json` tag)

But to answer your question, "1" and 1 and "1." do collide (but not true, true is/was "true")

I could prefix the key with the type, I'm actually doing that for a few types like error in the upcoming commit

ccoVeille: I'm sorry I edited my message after checking, but forgot one part.

Yes `true` doesn't collide with 1, but it does with `"true" `

So yes, also using the type could avoid collision.

extensions/extension.go
ldemailly: removed it

ccoVeille (approved):

extensions/extension.go
ccoVeille: The way you handled is elegant

ccoVeille:
Ah ! I was very surprised by theb

extensions/extension.go
ldemailly: Thanks! of note that yes the translation to json and then back isn't identity (at least not yet, unless I would implement custom deserialization) the current goal is more so people can produce json out of their grol data structures

main_pprof.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var (
+	cpuprofile = flag.String("profile-cpu", "", "write cpu profile to `file`")
+	memprofile = flag.String("profile-mem", "", "write memory profile to `file`")
+)
+
+func init() {
+	hookBefore = pprofBeforeHook
+	hookAfter = pprofAfterHook
```
ldemailly, line 21: hopefully that's ok in this case @ccoVeille (to have with and without pprof version without duplicating the whole main and without needing a main_no_pprof.go test

ccoVeille: This one is OK 👍

PR #117, opened 2024-08-06
Adding map.key as short for map["key"]

### LINKED ISSUE #118: dot access to map (`map.key` == `map["key"]`)
This will make access to things like `info.version` cleaner than `info["version"]` for string keys

Also will help for catch and other map based results/multi valued results (#105)

It also enables function namespaces (and OOish)

### LINKED ISSUE #119: env stack (`info.all_ids`) should be an array instead of a map with int indexes
Maybe the values in the array should be "name": "value", possible also a special value for name/line/...

ldemailly:
These`info["platform"]` were getting ugly, so now it's neat as `info.platform`

With this we can namespace functions easily

```go
ns1 = { 
   "foo": func(x) {x+1},
   "bar": func(x) {x-1}
}
ns1.foo(3)
```

- [x] fixed parser error: no prefix parse function for EOL found when running above interactively
- [x] fixed precedence so `ns.foo(3)` works without needing `(ns1.foo)(3)`

Closes #118 

Also implemented
- #119 

Closes #119


examples/sample.gr
```diff
@@ -28,7 +28,7 @@ m={"key": a, 73: 29} // so do maps
 println("m is:", m) // stdout print
 println("Outputting a smiley: 😀")
 
-first(m["key"]) // get the value from key from map, which is an array, and the first element of the array is our factorial 5
-// could also have been m["key"][0]
+first(m.key) // get the value from key from map, which is an array, and the first element of the array is our factorial 5
+// (dot notation .key is shorthand for ["key"]), could also have been firs(m["key"]) or m["key"][0] or m.key[0]
```
ccoVeille, line 32:
```suggestion
// (dot notation .key is shorthand for ["key"]), could also have been first(m["key"]) or m["key"][0] or m.key[0]
```

ccoVeille (approved):

PR #120, opened 2024-08-08
Initial integration with fortio.org/terminal

ldemailly:
Gets arrow keys and history - integration of several days of work to create https://github.com/fortio/terminal and improve the original golang.org/x/term through https://github.com/golang/term/pull/15 (and in a usable fork while waiting: https://github.com/fortio/term)

Only adds 51k 🎉 

```
┌─────────────────────────────────────────────────────────────────────┐
│ Diff between grol and grol_terminal                                 │
├──────────┬──────────────────────────┬──────────┬──────────┬─────────┤
│ PERCENT  │ NAME                     │ OLD SIZE │ NEW SIZE │ DIFF    │
├──────────┼──────────────────────────┼──────────┼──────────┼─────────┤
│ +100%    │ fortio.org/term          │          │ 24 kB    │ +24 kB  │
│ +100%    │ fortio.org/terminal      │          │ 5.9 kB   │ +5.9 kB │
│ +9.57%   │ <autogenerated>          │ 44 kB    │ 49 kB    │ +4.2 kB │
│ +340.55% │ path/filepath            │ 878 B    │ 3.9 kB   │ +3.0 kB │
│ +100%    │ golang.org/x/sys         │          │ 1.8 kB   │ +1.8 kB │
│ +33.03%  │ io                       │ 3.7 kB   │ 4.9 kB   │ +1.2 kB │
│ +7.70%   │ os                       │ 15 kB    │ 16 kB    │ +1.1 kB │
│ +6.59%   │ internal/poll            │ 14 kB    │ 16 kB    │ +958 B  │
│ +0.34%   │ grol.io/grol             │ 202 kB   │ 202 kB   │ +687 B  │
│ +0.09%   │ runtime                  │ 696 kB   │ 697 kB   │ +655 B  │
│ +2.95%   │ strings                  │ 20 kB    │ 21 kB    │ +591 B  │
│ +0.99%   │ fmt                      │ 51 kB    │ 52 kB    │ +511 B  │
│ +5.93%   │ main                     │ 4.6 kB   │ 4.9 kB   │ +272 B  │
├──────────┼──────────────────────────┼──────────┼──────────┼─────────┤
│ +3.04%   │ __rodata __DATA_CONST    │ 185 kB   │ 191 kB   │ +5.6 kB │
│ +1.78%   │ __rodata __TEXT          │ 196 kB   │ 199 kB   │ +3.5 kB │
│ +6.81%   │ __gopclntab __DATA_CONST │ 29 kB    │ 31 kB    │ +2.0 kB │
│ +3.05%   │ __data __DATA            │ 19 kB    │ 20 kB    │ +576 B  │
│ +24.19%  │ __go_buildinfo __DATA    │ 992 B    │ 1.2 kB   │ +240 B  │
│ +0.49%   │ __noptrdata __DATA       │ 32 kB    │ 33 kB    │ +160 B  │
│ +2.70%   │ __typelink __DATA_CONST  │ 2.2 kB   │ 2.3 kB   │ +60 B   │
│ +4.35%   │ __itablink __DATA_CONST  │ 552 B    │ 576 B    │ +24 B   │
├──────────┼──────────────────────────┼──────────┼──────────┼─────────┤
│ +2.50%   │ grol                     │ 2.0 MB   │ 2.1 MB   │ +51 kB  │
│          │ grol_terminal            │          │          │         │
└──────────┴──────────────────────────┴──────────┴──────────┴─────────┘
```


repl/repl.go
```diff
@@ -78,29 +84,77 @@ func EvalString(what string) (res string, errs []string, formatted string) {
...
+		return log.FErrf("Error creating readline: %v", err)
+	}
+	defer term.Close()
+	term.LoggerSetup()
+	term.SetPrompt(PROMPT)
+	options.Compact = true // because terminal doesn't do well will multi-line commands.
+	term.NewHistory(options.MaxHistory)
+	_ = term.SetHistoryFile(options.HistoryFile)
+	// Regular expression for "!nn" to run history command nn.
+	historyRegex := regexp.MustCompile(`^!(\d+)$`)
```
ccoVeille, line 105: Move this out or you will compute the regexp on each run

Also that's somehow the idea behind MustCompile

Otherwise use Compile and catch error

repl/repl.go
```diff
@@ -78,29 +84,77 @@ func EvalString(what string) (res string, errs []string, formatted string) {
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				// To avoid trailing prompt due to prompt refresh on the log output
+				// that's a bit ugly that it's necessary. consider handling in terminal.Close
+				term.SetPrompt("")
+				log.Infof("Exit requested") // Don't say EOF as ^C comes through as EOF as well.
+				return 0
+			}
+			return log.FErrf("Error reading line: %v", err)
+		}
```
ccoVeille, line 117:
```suggestion
		if errors.Is(err, io.EOF) {
			// To avoid trailing prompt due to prompt refresh on the log output
			// that's a bit ugly that it's necessary. consider handling in terminal.Close
			term.SetPrompt("")
			log.Infof("Exit requested") // Don't say EOF as ^C comes through as EOF as well.
			return 0
		}
		if err != nil {
			return log.FErrf("Error reading line: %v", err)
		}
```

main.go
```diff
@@ -28,24 +29,34 @@ func Main() int {
 	compact := flag.Bool("compact", false, "When printing code, use no indentation and most compact form")
 	showEval := flag.Bool("eval", true, "show eval results")
 	sharedState := flag.Bool("shared-state", false, "All files share same interpreter state (default is new state for each)")
+	homeDir, err := os.UserHomeDir()
+	historyFileDefault := filepath.Join(homeDir, ".grol_history")
```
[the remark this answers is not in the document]
ldemailly: ah I forgot os.UserConfigDir existed - even though personally I don't want my files in unfindable `$HOME/Library/Application Support` but... sure

I also don't want to make new dir though

repl/repl.go
ldemailly: there is only exactly 1 run

main.go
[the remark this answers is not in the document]
ccoVeille: What about adding GROL_HISTORY env variable, that will be used instead of os.UserConfDir when found?

repl/repl.go
ldemailly: thanks for the reminder, that code moved to terminal/ so isn't needed at all anymore

main.go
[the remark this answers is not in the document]
ldemailly: I like that better though it bothers me to have to define anything to not get this horror
```
  -history file
    	history file to use (default "/Users/dl/Library/Application Support/.grol_history")
```

main.go
```diff
@@ -28,24 +29,34 @@ func Main() int {
 	compact := flag.Bool("compact", false, "When printing code, use no indentation and most compact form")
 	showEval := flag.Bool("eval", true, "show eval results")
 	sharedState := flag.Bool("shared-state", false, "All files share same interpreter state (default is new state for each)")
+	configDir, err := os.UserConfigDir() // TODO: this is ugly on a mac... allow homedir...
+	historyFileDefault := filepath.Join(configDir, ".grol_history")
```
ccoVeille, line 33:
```suggestion
	historyFileDefault := filepath.Join(configDir, "grol", "history")
```

ldemailly: and now I have to mkdir all etc...? isn't it _enough_ to use the config dir and put files there

ccoVeille: It's dirty. I would prefer if you added a .grol_history file at the root of home folder than creating a file at the root of config folder.

ccoVeille (approved):

main.go
ldemailly: ok so I switched back to homedir as default but added GROL_HISTORY_FILE env var for people who don't like it so they don't have to set it in the flag either, does that work?

main.go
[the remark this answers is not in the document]
ldemailly: see latest

PR #121, opened 2024-08-10
Adding a Trie, using it for completion

ldemailly:
Fairly optimized:

- use [256] arrays which are accessed without bound check
- use a shared marker for the leaves to save memory on all leaves

Use it to auto complete by prefix on tab

Keep adding the defined IDs/Functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,12 @@ $ info.keywords
 == Eval  ==> ["else","error","false","first","func","if","len","log","macro","print","println","quote","rest","return","true","unquote"]
 ```
 
-There is also in interactive repl mode: `history`, `!23` to repeat the 23rd statement for instance and `help`.
-And full edit and history navigation with arrow keys etc... and hit the `<tab>` key at anytime for completion.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ $ info.keywords
 ```
 
 There is also in interactive repl mode: `history`, `!23` to repeat the 23rd statement for instance and `help`.
-And full edit and history navigation with arrow keys etc...
+And full edit and history navigation with arrow keys etc... and hit the `<tab>` key at anytime for completion.
```
ccoVeille, line 69: Two and in the same sentence.

It sounds strange, especially when previous one has an and

You could use a list like this

 There is also in interactive repl mode:
- `history`
- `!23` to repeat the 23rd statement for instance
- `help`. 
- full edit and history navigation with arrow keys
-  hit the `<tab>` key at anytime for completion.
- etc...

ldemailly: thanks, good suggestion! done

PR #122, opened 2024-08-11
Adding load(fname) and save(fname) - save being really eval_file(fname)

### LINKED ISSUE #38: add save/load

### LINKED ISSUE #125: can't parse large numbers
```go
$ x=1e19
10000000000000000000
$ x=10000000000000000000
14:23:43.887 [CRI] parser has 1 error(s)
14:23:43.887 [ERR] parser error: could not parse "10000000000000000000" as integer
```

ldemailly:
Fixes #38 
Fixes #125 

tests have been manual, loading all the examples, saving, reloading comparing (how I stepped on #125) 
not exactly sure how to ... actually test this

constants if not exact might cause an issue

### DIFF
--- a/eval/macro_expension.go
+++ b/eval/macro_expension.go
@@ -9,11 +9,11 @@ import (
 	"grol.io/grol/token"
 )
 
-func DefineMacros(store *object.Environment, program *ast.Statements) {
+func (s *State) DefineMacros(program *ast.Statements) {
 	for i := 0; i < len(program.Statements); /* not always incrementing */ {
 		statement := program.Statements[i]
 		if isMacroDefinition(statement) {
-			addMacro(store, statement)
+			addMacro(s.macroState, statement)
 			program.Statements = append(program.Statements[:i], program.Statements[i+1:]...)
 		} else {
 			i++
@@ -72,14 +72,14 @@ func isMacroCall(s *object.Environment, exp *ast.CallExpression) (*object.Macro,
 	return macro, true
 }
 
-func ExpandMacros(s *object.Environment, program ast.Node) ast.Node {
+func (s *State) ExpandMacros(program ast.Node) ast.Node {
 	return ast.Modify(program, func(node ast.Node) ast.Node {
 		callExpression, ok := node.(*ast.CallExpression)
 		if !ok {
 			return node
 		}
 
-		macro, ok := isMacroCall(s, callExpression)
+		macro, ok := isMacroCall(s.macroState, callExpression)
 		if !ok {
 			return node
 		}
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -4,9 +4,12 @@ package extensions
 
 import (
 	"fmt"
+	"io"
 	"math"
+	"os"
 	"strings"
 
+	"fortio.org/log"
 	"grol.io/grol/eval"
 	"grol.io/grol/object"
 )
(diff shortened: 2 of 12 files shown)


ldemailly:
because I moved some code from one file (eval.go) to another (eval_api.go) it's probably easier to review individual commits

eval/macro_expension.go
```diff
@@ -9,11 +9,11 @@ import (
 	"grol.io/grol/token"
 )
 
-func DefineMacros(store *object.Environment, program *ast.Statements) {
+func (s *State) DefineMacros(program *ast.Statements) {
```
ccoVeille, line 12: Do you think they have to be exported?

eval/macro_expension.go
```diff
@@ -72,14 +72,14 @@ func isMacroCall(s *object.Environment, exp *ast.CallExpression) (*object.Macro,
 	return macro, true
 }
 
-func ExpandMacros(s *object.Environment, program ast.Node) ast.Node {
+func (s *State) ExpandMacros(program ast.Node) ast.Node {
```
ccoVeille, line 75: Same question

extensions/extension.go
```diff
@@ -158,3 +173,42 @@ func evalFunc(env any, name string, args []object.Object) object.Object {
 	}
 	return res
 }
+
+func saveFunc(env any, _ string, args []object.Object) object.Object {
```
ccoVeille, line 177: Does this method must comply with an interface?

You should add a comment then, because the unused string is awkward

extensions/extension.go
```diff
@@ -158,3 +173,42 @@ func evalFunc(env any, name string, args []object.Object) object.Object {
...
+	// Write to file.
+	n, err := eval.SaveGlobals(f)
+	if err != nil {
+		return object.Error{Value: err.Error()}
+	}
+	log.Infof("Saved %d ids/fns to: %s", n, file)
+	return object.NULL
+}
+
+func loadFunc(env any, _ string, args []object.Object) object.Object {
```
ccoVeille, line 195: Same remark

object/state.go
```diff
@@ -94,6 +96,43 @@ func (e *Environment) RegisterTrie(t *trie.Trie) {
...
+	for e.outer != nil {
+		e = e.outer
+	}
+	keys := make([]string, 0, len(e.store))
+	for k := range e.store {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+	n := 0
+	for _, k := range keys {
```
ccoVeille, line 110: I might be missing something but either you add a comment explaining why you need `n`

or you do that


```suggestion
	for n, k := range keys {
```

Then return n+1 at the end of the function

eval/macro_expension.go
ldemailly: yes it's used from repl which isn't in eval

object/state.go
ldemailly: n would be out of scope...

ldemailly: I can do `var k string` though and drop the `:=` though it seems icky to use loop vars outside of the loop

ldemailly: but that'd be incorrect for empty keys

extensions/extension.go
ldemailly: yeah it's defined a bit above

```go
	jsonFn.Callback = saveFunc // save to file.
```
which is
```go
// ExtFunction is the signature of what grol will call when the extension is invoked.
// Incoming arguments are validated for type and number of arguments based on [Extension].
// eval is the opaque state passed from the interpreter, it can be used with eval.Eval etc
// name is the function name as registered under.
type ExtFunction func(eval any, name string, args []Object) Object
```

like everything in the file, in short it doesn't need to know its own name

object/state.go
ldemailly: so in short n counts the number of succcesful writes, it happens to be len(keys) if there are no errors but can be shorter if there is an error.

PR #123, opened 2024-08-11
[do not merge] Std wasm

ldemailly:
Just to quickly apply/compare wasm with stdgo'

Should probably make it a makefile variable


ldemailly:
applied this actually

PR #127, opened 2024-08-12
restrict by default, avoid security issues, force files to be alphanum.gr

### LINKED ISSUE #126: have a safe way to disable or restrict load/save (in particular for discord bot)
either restricted to a given directory and/or completely disabled

ldemailly:
following options are added (with safe defaults)
```go
type ExtensionConfig struct {
	HasLoad           bool
	HasSave           bool
	LoadSaveEmptyOnly bool
	UnrestrictedIOs   bool // Dangerous, can overrwite files, read any readable file etc...
}
```

Fixes #126

### DIFF
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -11,6 +11,7 @@ import (
 
 	"fortio.org/log"
 	"grol.io/grol/eval"
+	"grol.io/grol/lexer"
 	"grol.io/grol/object"
 )
 
@@ -19,18 +20,22 @@ var (
 	errInInit error
 )
 
-func Init() error {
+type ExtensionConfig struct {
+	LoadAndSave bool
+}
+
+func Init(c *ExtensionConfig) error {
 	if initDone {
 		return errInInit
 	}
-	errInInit = initInternal()
+	errInInit = initInternal(c)
 	initDone = true
 	return errInInit
 }
 
 type OneFloatInOutFunc func(float64) float64
 
-func initInternal() error {
+func initInternal(c *ExtensionConfig) error {
 	cmd := object.Extension{
 		Name:     "pow",
 		MinArgs:  2,
@@ -127,6 +132,9 @@ func initInternal() error {
 	if err != nil {
 		return err
 	}
+	if !c.LoadAndSave {
+		return nil
+	}
 	jsonFn.Name = "save"
 	jsonFn.Callback = saveFunc // save to file.
 	err = object.CreateFunction(jsonFn)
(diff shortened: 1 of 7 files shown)


extensions/extension.go
```diff
@@ -174,10 +202,36 @@ func evalFunc(env any, name string, args []object.Object) object.Object {
...
+		return file, nil
+	}
+	// only alhpanumeric and _ allowed. no dots, slashes, etc.
+	f := strings.TrimSuffix(file, ".gr")
+	for _, r := range []byte(f) {
+		if !lexer.IsAlphaNum(r) {
+			return "", fmt.Errorf("invalid character in filename %q: %c", file, r)
+		}
+	}
+	return f + ".gr", nil
```
ccoVeille, line 221: Why not simply returning


```suggestion
	return file, nil
```

extensions/extension.go
```diff
@@ -174,10 +202,36 @@ func evalFunc(env any, name string, args []object.Object) object.Object {
...
+	file := ".gr"
+	if len(args) != 0 {
+		file = args[0].(object.String).Value
+		var err error
+		file, err = sanitizeFileName(file)
+		if err != nil {
+			return object.Error{Value: err.Error()}
+		}
+	}
 	f, err := os.Create(file)
```
ccoVeille, line 235: So it could will create a file named .gr if there is no art?

extensions/extension.go
```diff
@@ -193,8 +247,15 @@ func saveFunc(env any, _ string, args []object.Object) object.Object {
...
-	file := args[0].(object.String).Value
-	// Open file for reading.
+	file := ".gr"
+	if len(args) != 0 {
+		file = args[0].(object.String).Value
+		var err error
+		file, err = sanitizeFileName(file)
+		if err != nil {
+			return object.Error{Value: err.Error()}
+		}
```
ccoVeille, line 257: As the code is duplicated between save and load, shouldn't it be moved into sanitizeFilename, by passing a []object.Object as argument

extensions/extension.go
ldemailly: because .gr is optional in the input (ie loads/save("foo") is valid (and will end up using foo.gr))

extensions/extension.go
ldemailly: That's what "EmptyOnly" is for, allows only 1 file called .gr

extensions/extension.go
ldemailly: fixed, thanks https://github.com/grol-io/grol/pull/130/files

extensions/extension.go
ccoVeille: Indeed, then. You trim if present, then add it

PR #139, opened 2024-08-14
Fixed caching of errors bug, new `-max-depth` and consistent panic recovery, autoload continue on errors (+Inf...), abandon tinygo, prevent NaN in maps, == works on more types

### LINKED ISSUE #137: == on functions
should work using cachkey as that's what we use anyway for memoization

then might as well allow it for map keys

### LINKED ISSUE #138: floats like +Inf and NaN can't be re-read
possible hack, predefine them

and add `+` as prefix so

nan=1/.0

serialized to

nan=+Inf

works

ldemailly:
- implement `-max-depth` and panic recovery in all Eval variants so infinite recursion, 1/0, etc. are caught
- fixed bug with caching of errors
- make autoload continue on errors (+Inf...) helps with #138 
- abandoning tinygo because https://tinygo.org/docs/reference/lang-support/#recover-builtin is the last straw that broke my back.
- fixed #138 
- fixed  #137


eval/eval_test.go
```diff
@@ -852,3 +853,35 @@ func TestExtension(t *testing.T) {
...
+
+func TestNaNMapKey(t *testing.T) {
+	s := eval.NewState()
+	_, err := eval.EvalString(s, `nan=0./0.`, false)
+	if err != nil {
+		t.Errorf("should have not errored out just defining NaN, got %v", err)
+	}
+	_, errs, _ := repl.EvalString(`nan=0./0; m={-42.3: "about -42", 42.1:"42.1", nan: "this is NaN"}; println(m)`)
+	if len(errs) != 1 {
+		t.Errorf("should have an error trying to put a nan in map, got %v", errs)
```
ccoVeille, line 881:
```suggestion
		t.Errorf("should have an error trying to put a NaN in map, got %v", errs)
```

Unless you have something against uppercase in log (but I don't think so, as other error are using NaN

Or maybe you love naan, and made a typo 🤣😂

main.go
```diff
@@ -100,7 +102,8 @@ func Main() int {
 		return log.FErrf("Error initializing extensions: %v", err)
 	}
 	if *commandFlag != "" {
-		res, errs, _ := repl.EvalString(*commandFlag)
+		// return EvalStringWithOption(Options{All: true, ShowEval: true, NoColor: true, Compact: false}, what)
```
ccoVeille, line 105: Left behind debug?

repl/repl.go
```diff
@@ -71,15 +75,30 @@ func AutoLoad(s *eval.State, options Options) error {
...
+		line := scanner.Text()
+		_, err = eval.EvalString(s, line, false)
+		if err == nil {
+			count++
+		} else {
+			errorCount++
+			errs = append(errs, err)
+			log.Errf("Error loading autoload line %q: %v", line, err)
+		}
 	}
```
ccoVeille, line 94:
```suggestion
	for scanner.Scan() {
		line := scanner.Text()
		_, err = eval.EvalString(s, line, false)
		if err == nil {
			count++
			continue
		}

		errorCount++
		errs = append(errs, err)
		log.Errf("Error loading autoload line %q: %v", line, err)
 	}
```

repl/repl.go
```diff
@@ -140,29 +159,28 @@ func EvalString(what string) (string, []string, string) {
...
 	out := &strings.Builder{}
 	s.Out = out
 	s.LogOut = out
 	s.NoLog = true
-	err := AutoLoad(s, o)
-	if err != nil {
-		log.Errf("Error loading autoload file: %v", err)
-	}
-	_, errs, formatted = EvalOne(s, what, out, o)
+	_ = AutoLoad(s, o) // errors already logged
```
ccoVeille, line 170: Why returning an error then?

repl/repl.go
```diff
@@ -232,20 +247,24 @@ func Interactive(options Options) int {
...
 				// In addition to raw lines, we also add the single line version to history.
 				log.LogVf("Adding to history: %q", formatted)
 				term.AddToHistory(formatted)
 			}
 			prev = ""
 			term.SetPrompt(PROMPT)
 		}
+		if panicked {
+			log.Errf("%s", errs[0]) // we know there is exactly 1 error in this case, see EvalOne defer.
+		}
```
ccoVeille, line 267: I don't get the need for this log as the defer recover already log the error

wasm/wasm_main.go
```diff
@@ -40,7 +51,11 @@ func jsEval(this js.Value, args []js.Value) interface{} {
 		anyErrs[i] = v
 	}
 	result["errors"] = anyErrs
-	result["formatted"] = strings.TrimSuffix(formatted, "\n")
+	fmt := strings.TrimSuffix(formatted, "\n")
```
ccoVeille, line 54: fmt is a package name, I wouldn't shadow it

ccoVeille (commented):
With love, Christophe

eval/eval.go
```diff
@@ -247,21 +251,21 @@ func (s *State) evalPrintLogError(node *ast.Builtin) object.Object {
 func (s *State) evalBuiltin(node *ast.Builtin) object.Object {
 	// all take 1 arg exactly except print and log which take 1+.
 	t := node.Type()
-	min := 1
+	minV := 1
```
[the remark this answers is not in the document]
ldemailly: yeah that was a new annoying warning:)

README.md
```diff
@@ -195,6 +195,8 @@ flags:
     	don't execute, just parse and re format the input
   -history file
     	history file to use (default "~/.grol_history")
+  -max-depth int
+    	Maximum interpreter depth (default 100000)
```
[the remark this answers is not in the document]
ldemailly: ah yes I increased it because of pi2.gr and didn't update the readme, I probably need to automate updating that section like I do for fortio

eval/eval_api.go
```diff
@@ -78,7 +87,14 @@ func (s *State) UpdateNumSet() (oldvalue, newvalue int64) {
 
 // Does unwrap (so stop bubbling up) return values.
 func (s *State) Eval(node any) object.Object {
+	if s.depth > s.MaxDepth {
+		log.LogVf("max depth %d reached", s.MaxDepth) // will be logged by the panic handler.
```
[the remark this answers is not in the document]
ldemailly: it was log.Critf for a while but it's also log critf there so here, closer to where it happens it's logged below the default log level

eval/eval_test.go
ldemailly: I like naan but yes should be NaN here

wasm/wasm_main.go
ccoVeille: I recommend you enabling "import-shadowing" linter in revive.

Available via golangci-lint

ldemailly: if you remember I have enable-all already so... it should show up

ldemailly: oh so it's a non default rule of revive?

ldemailly: not sure how to add 1 extra without having to specify all the other default, running it by hand once with:
```
linters-settings:
  revive:
    rules:
      - name: import-shadowing
```

ldemailly: funny it found a few others (like errors and eval) but it doesn't mind this one :-)

eval/eval.go
```diff
@@ -405,6 +409,11 @@ func (s *State) applyFunction(name string, fn object.Object, args []object.Objec
 	s.Out = oldOut
 	output := buf.Bytes()
 	_, _ = s.Out.Write(output)
```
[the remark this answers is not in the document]
ldemailly: not really but then I do in print log a warning (if there are errors writing to Out the logger will likely also fail to write)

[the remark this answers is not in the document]
ldemailly: added

README.md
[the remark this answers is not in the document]
ldemailly: fixed

repl/repl.go
ldemailly: well... still.. maybe there would be a strict mode and you would stop on errors

repl/repl.go
ldemailly: you're right, I think it was something to do with showing up on same line as eval but... removed

before
```
$ 0/0
== Parse ==> 0/0
== Eval  ==> 09:07:43.886 [CRI] Caught panic: runtime error: integer divide by zero
09:07:43.886 [ERR] panic: runtime error: integer divide by zero
```

now
```
$ func (){self()}()
== Parse ==> func(){self()}()
== Eval  ==> 09:10:16.289 [CRI] Caught panic: max depth 250000 reached
```

(that's with -parse, without it looks ok)

main.go
ldemailly: yeah or rather I copied that one from EvalString to see which options were set before

wasm/wasm_main.go
ccoVeille: You can find the revive default ones in my 03-safe example here

https://github.com/ccoVeille/golangci-lint-config-examples

I have enabled them all and it helped me.

I still have to share a conf about it

eval/eval_api.go
```diff
@@ -78,7 +87,14 @@ func (s *State) UpdateNumSet() (oldvalue, newvalue int64) {
...
+		s.depth = 0
```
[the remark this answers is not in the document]
ldemailly: ah I forgot this, somehow not showing in main thread, adding comment now in next PR

PR #140, opened 2024-08-14
Fix bug with blankslate having 0 max depth recursion

ldemailly:
(also adding comment missing from #139 reviews)

### DIFF
--- a/eval/eval_api.go
+++ b/eval/eval_api.go
@@ -52,6 +52,7 @@ func NewBlankState() *State {
 		cache:      NewCache(),
 		extensions: make(map[string]object.Extension),
 		macroState: object.NewMacroEnvironment(),
+		MaxDepth:   DefaultMaxDepth,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


eval/eval_api.go
```diff
@@ -125,12 +127,15 @@ func EvalString(this any, code string, emptyEnv bool) (object.Object, error) {
 	if len(p.Errors()) != 0 {
 		return object.NULL, fmt.Errorf("parsing error: %v", p.Errors())
 	}
-	var evalState *State
+	evalState, ok := this.(*State)
```
ccoVeille, line 130: I'm only noticing it now, but a variable named this/self looks odd in Go code.

ldemailly: why ? it is "this" in this case (ie the eval state passed back through an any arg)

ccoVeille: this like self looks like a reference to current object, as in PHP or JavaScript

ccoVeille: But I might be the only one having nightmare about these languages

ldemailly: yet.. that's what it is here (*State) is the receiver in the rest of the package thus the name

PR #142, opened 2024-08-14
Big change to grol maps, use sorted slice of kv underneath instead of actual go map

ldemailly:


object/object.go
```diff
@@ -53,6 +53,7 @@ type Number interface {
 }
 */
 
+// Hashable in tem of go map for cache key.
```
ccoVeille, line 56:
```suggestion
// Hashable in tem of Go map for cache key.
```

object/object.go
```diff
@@ -71,38 +72,105 @@ func NativeBoolToBooleanObject(input bool) Boolean {
...
+	// RETURN, QUOTE, MACRO, ANY aren't expected to be compared.
+	case RETURN:
+		fallthrough
+	case QUOTE:
+		fallthrough
+	case MACRO:
+		fallthrough
+	case UNKNOWN:
+		fallthrough
+	case ANY:
```
ccoVeille, line 170: The fallthrought makes sense when you have


```go
	case RETURN:
		fmt.Println...
		fallthrough
	case QUOTE:
		callSonething()
		fallthrough
	case MACRO:
		i++
		fallthrough
	case UNKNOWN:
		i++
		fallthrough
	case ANY:
		logSomething
```

Here what you want to do should be like this

```suggestion
	case RETURN, QUOTE, MACRO, UNKNOWN, ANY:
```

Or (only a stylistic variation)

```suggestion
	case
		RETURN,
		QUOTE,
		MACRO,
		UNKNOWN,
		ANY:
```

object/object.go
```diff
@@ -117,18 +185,63 @@ func ArrayEquals(left, right []Object) bool {
...
 	return true
 }
 
+func CompareKeys(a, b KV) int {
+	return Cmp(a.Key, b.Key)
+}
+
+func (m *Map) Get(key Object) (Object, bool) {
+	kv := KV{Key: key}
+	i, ok := slices.BinarySearchFunc(m.kv, kv, CompareKeys)
```
ccoVeille, line 209: Maybe a comment explaining why a binarysearch is important, it's not a well-known method

object/object.go
```diff
@@ -338,108 +451,76 @@ func (ao Array) JSON(w io.Writer) error {
 
 // possible optimization: us a map of any and put the inner value of object in there, would be faster than
 // wrapping strings etc into object.
-type Map map[Object]Object
+type KV struct {
```
ccoVeille, line 454: I know what KV state for, but KeyValue would be clearer for everyone

object/object.go
ldemailly: there should be a linter for that :) found 2 spots, fixing

object/object.go
ldemailly: considering not exporting it also as it's barely used outside

object/object.go
ldemailly: fixed in #147

PR #143, opened 2024-08-15
first/rest for strings (rune) and maps;  str[index] is the byte value; implement keys(map) in grol; fix save's usage (1 optional arg case)

ldemailly:
- first/rest for strings (rune) and maps; for maps it returns a map with .key and .value keys
- str[index] is the byte value
- implement keys(map) in grol
```go
func keys(m) {
    if (len(m)==0) {
        return []
    }
    return [first(m).key]+keys(rest(m))
}
```
- fix save's usage (1 optional arg case)

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -231,6 +231,29 @@ func (m *Map) KV() []KV {
 	return m.kv
 }
 
+var (
+	KeyKey   = String{Value: "key"}
+	ValueKey = String{Value: "value"}
+)
+
+func (m *Map) First() Object {
+	if len(m.kv) == 0 {
+		return NULL
+	}
+	return &Map{kv: []KV{
+		{Key: KeyKey, Value: m.kv[0].Key},
+		{Key: ValueKey, Value: m.kv[0].Value},
+	}}
+}
+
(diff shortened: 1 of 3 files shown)


object/object.go
```diff
@@ -231,6 +231,29 @@ func (m *Map) KV() []KV {
 	return m.kv
 }
 
+var (
+	KeyKey   = String{Value: "key"}
+	ValueKey = String{Value: "value"}
```
ccoVeille, line 236:
```suggestion
	Key   = String{Value: "key"}
	Value = String{Value: "value"}
```

Or a common prefix


```suggestion
	KeyKey   = String{Value: "key"}
	KeyValue = String{Value: "value"}
```

ldemailly: can't use the first one (that was what I had) because it's same name as the fields (well you can but it looks weird to have Key: Key, though that's what I wrote at first)

ldemailly: and it's keys so it's the value's key, I think a suffix is clear ValueKey vs KeyValue sounds like a Value

ccoVeille: OK 👍

PR #147, opened 2024-08-15
Get rid of regexp, save 250kb on binary size, tweak install, fix bug with !n history off by one when full, nil/NaN/Inf ids, map append fix, array[nil]==array[0]

### LINKED ISSUE #144: drop regexp
it's costing 120kb
│ +100%    │ regexp                         │          │ 120 kB   │ +120 kB │

to do

```go
repl/repl.go:   historyRegex := regexp.MustCompile(`^!(\d+)$`)
```

### LINKED ISSUE #145: !n is off by one when history is full
because !xx is added to the history and thus shifting it all

### LINKED ISSUE #148: consider x[nil] same as x[0]
would allow to write:
```go
func explode(s, ..) {
	if len(..) == 0 {
		i = 0
	} else {
		i = ..[0]
	}
	if i >= len(s) {
		return []
	}
	return [(s[i])] + self(s, i + 1)
}
s = "ABC"
explode(s)
```
without the first if

ldemailly:
- #144 Get rid of regexp save 250kb on binary size
- Tweak install, add `make install`
- #145 Fix bug with !n history off by one when full
- Add nil,NaN,Inf identifiers
- Allow nil[whichever] -> nil
- Add unary + (prefix + == noop)
- #148 
- Fix bug where append (+) on map with same key would mutate the original

fixes #144 
fixes #145 
fixes #148

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"regexp"
 	"runtime/debug"
 	"slices"
 	"strconv"
@@ -177,6 +176,16 @@ func EvalStringWithOption(o Options, what string) (res string, errs []string, fo
 	return
 }
 
+func extractHistoryNumber(input string) (int, bool) {
+	if len(input) > 1 && input[0] == '!' {
+		numberPart := input[1:]
+		if num, err := strconv.Atoi(numberPart); err == nil {
+			return num, true
+		}
+	}
+	return 0, false
+}
+
 func Interactive(options Options) int {
 	options.NilAndErr = true
 	s := eval.NewState()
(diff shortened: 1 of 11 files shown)


repl/repl.go
```diff
@@ -177,6 +176,16 @@ func EvalStringWithOption(o Options, what string) (res string, errs []string, fo
 	return
 }
 
+func extractHistoryNumber(input string) (int, bool) {
+	if len(input) > 1 && input[0] == '!' {
+		numberPart := input[1:]
```
ccoVeille, line 181: Have you tried looking at strings.Cut ? And its variant CutPrefix?

https://pkg.go.dev/strings#Cut
https://pkg.go.dev/strings#CutPrefix

ldemailly: well, if my prefix was not a single byte, sure but I don't think it'd help much here ? (not that it matters for this but this one will perform better, and also I don't think it's any less clear? only benefit of CutPrefix would be... to advertise it exists (which I didn't I will admit))

PR #159, opened 2024-08-18
Solve/avoid OOM kills and memory fatals by doing our own checks

ldemailly:
so

```
GOMEMLIMIT=10GiB ./grol

$ len([1,2,3]*200_000_000)
600000000
$ len([1,2,3]*300_000_000)
10:59:22.702 [CRI] Caught panic: would exceed memory requesting 900000000 objects, 10736636880 free
$ len("ABC"*3_000_000_000)
9000000000
$ len("ABC"*4_000_000_000)
10:58:09.054 [CRI] Caught panic: would exceed memory requesting 750000000 objects, 10736643576 free
```

also to limit IOs:
```
$ long=["ABC"]*1000
[...]
11:22:00.014 [INF] Exit requested
11:22:00.016 [WRN] Skipping "long" as it's too long (6001 > 4000)
11:22:00.016 [INF] Auto saved 30 ids/fns (1 set) to: .gr
```

this originates from stepping on https://github.com/ldemailly/go-scratch/blob/main/membug/membug.go ie you can allocate 10x the limit and get no error, until you get OOM killed _or_ fatal, but no recoverable error, this makes it recoverable

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Use `info` to see all the available functions, keywords, operators etc... (can b
 
 See also [sample.gr](examples/sample.gr) and others in that folder, that you can run with
 ```
-gorepl examples/*.gr
+grol examples/*.gr
 ```
 
 or copypaste to the online version on [grol.io](https://grol.io)
--- a/check_samples_double_format.sh
+++ b/check_samples_double_format.sh
@@ -1,5 +1,6 @@
 #! /bin/bash
 set -e
+export GOMEMLIMIT=1GiB
 for file in "$@"; do
     echo "---testing double format for $file---"
     ./grol -format $file > /tmp/format1
--- a/main.go
+++ b/main.go
@@ -5,8 +5,10 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
+	"runtime/debug"
 
 	"fortio.org/cli"
 	"fortio.org/log"
@@ -74,6 +76,10 @@ func Main() int {
 		}
 	}
 	log.Infof("grol %s - welcome!", cli.LongVersion)
+	memlimit := debug.SetMemoryLimit(-1)
+	if memlimit == math.MaxInt64 {
+		log.Warnf("Memory limit not set, please set GOMEMLIMIT=1GiB or similar")
+	}
 	options := repl.Options{
 		ShowParse:   *showParse,
 		ShowEval:    *showEval,
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func Main() int {
 	log.Infof("grol %s - welcome!", cli.LongVersion)
 	memlimit := debug.SetMemoryLimit(-1)
 	if memlimit == math.MaxInt64 {
-		log.Warnf("Memory limit not set, please set GOMEMLIMIT=1GiB or similar")
+		log.Warnf("Memory limit not set, please set the GOMEMLIMIT env var; e.g. GOMEMLIMIT=1GiB")
 	}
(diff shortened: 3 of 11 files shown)


main.go
```diff
@@ -74,6 +77,10 @@ func Main() int {
 		}
 	}
 	log.Infof("grol %s - welcome!", cli.LongVersion)
+	memlimit := debug.SetMemoryLimit(-1)
```
ccoVeille, line 80: A comment about the -1 would be appreciated, it's a magical number

main.go
```diff
@@ -74,6 +77,10 @@ func Main() int {
...
+	if memlimit == math.MaxInt64 {
+		log.Warnf("Memory limit not set, please set GOMEMLIMIT=1GiB or similar")
```
ccoVeille, line 82: Maybe add something about the fact you are talking about an environment variable

ldemailly:
will use in grol-io/grol-discord-bot#80 to avoid bot crashes

main.go
ldemailly: it's in the  doc https://pkg.go.dev/runtime/debug#SetMemoryLimit but will add

> A negative input does not adjust the limit, and allows for retrieval of the currently set memory limit.

main.go
ldemailly: changing to
```
12:07:34.993 [WRN] Memory limit not set, please set the GOMEMLIMIT env var; e.g. GOMEMLIMIT=1GiB
```

README.md
```diff
@@ -97,7 +97,7 @@ Use `info` to see all the available functions, keywords, operators etc... (can b
 
 See also [sample.gr](examples/sample.gr) and others in that folder, that you can run with
 ```
```
ccoVeille, line 99:
````suggestion
```shell
````

File: check_samples_double_format.sh [hunk 4]
```diff
@@ -1,5 +1,6 @@
 #! /bin/bash
```
ccoVeille, line 1: Do you use shfmt and shell check?

https://github.com/mvdan/sh

https://github.com/koalaman/shellcheck

There are GHA for them, they are pretty useful

main.go
ccoVeille: I know it's in the doc, but it's a bit cryptic when you don't know that you call set with -1 to get the value.

Thanks for adding the comment

check_samples_double_format.sh
ldemailly: I use shellcheck for work, haven't here

StevenACoffman:
Hey, I have put this PR in a place of honor in a number of my "tips and tricks" guides. People are usually quite impressed!

I know this is old, and you are likely already aware, but I usually mention this PR in the same breath as pointing out the pitfalls of running inside a container where Go can sometimes accidentally confuse the _**container's**_ resource limits with the **_host's_** resource limits. 

There are some band-aid library workarounds:

https://github.com/KimMachineGun/automemlimit
https://github.com/uber-go/automaxprocs

ldemailly:
@StevenACoffman glad you find it useful (where is that tips and tricks guide btw?)

Personally I think CPU limits are _bad_ which is also why I think automaxprocs is also bad: they waste precious sys cpu for unecessary enforcement, as cpu _is_ elastic (unlike memory) 

https://github.com/golang/go/issues/73193#issuecomment-2784715261
and my comment at the bottom of https://blog.howardjohn.info/posts/gomaxprocs/

But this is about memory, which is definitively not elastic

ldemailly:
https://github.com/golang/go/issues/68934 is the (sadly closed) go issue, maybe new leadership will reconsider

PR #163, opened 2024-08-19
Introduce SmallArray to cache smaller arrays in memoization cache. new `-panic` option for development/debug

ldemailly:
Also cleans up the code by making the internal object.Array Elements private.


ccoVeille (commented):
A lot of changes, but it seems good

ldemailly:
thx, considering adding "small map" too

PR #164, opened 2024-08-19
Small maps

ldemailly:
Similar to #163 

So small maps (up to 4 keys) can be cached at expense of copying

Renamed the previous object.Map -> object.BigMap
Made a Map interface (called MapInterface during development and back to using Map now after checking references to BigMap etc)

### DIFF
--- a/eval/eval_test.go
+++ b/eval/eval_test.go
@@ -580,7 +580,7 @@ func TestMapLiterals(t *testing.T) {
     }`
 
 	evaluated := testEval(t, input)
-	result, ok := evaluated.(*object.Map)
+	result, ok := evaluated.(object.Map)
 	if !ok {
 		t.Fatalf("Eval didn't return Map. got=%T (%+v)", evaluated, evaluated)
 	}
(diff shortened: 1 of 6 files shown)


ldemailly:
found a bug in Cmp while doing self review, need also to fix for Array

ldemailly:
(not exactly great) perf test as it felt slower to run `make` 2 runs after first one on this branch
```
echo "" | make
       10.85 real        10.68 user         4.81 sys
       10.71 real        10.66 user         4.91 sys
```
on main
```
       10.80 real        10.61 user         5.06 sys
       10.59 real        10.67 user         5.04 sys
```
sys is lower most of time is in pi2.gr

ldemailly:
Did a bunch of testing by hand but... need
- #165

eval/eval_test.go
```diff
@@ -931,6 +931,20 @@ func TestMapAccidentalMutation(t *testing.T) {
...
+
+func TestSmallMapSorting(t *testing.T) {
+	inp := `m={2:"b"};n={1:"a"};m+n`
+	expected := `{1:"a",2:"b"}`
+	s := eval.NewState()
+	res, err := eval.EvalString(s, inp, false)
+	if err != nil {
+		t.Errorf("should not have errored: %v", err)
+	}
+	resStr := res.Inspect()
```
ccoVeille, line 946: Won't it panic on error?

I mean isn't res=nil on error?

What about this?


```suggestion
	if err != nil {
		t.Errorf("should not have errored: %v", err)
		return
	}
	resStr := res.Inspect()
```

ldemailly: I don't think it does no, I think you get object.NULL but .. even if it does and the test crash, it's ok (agreed through that t.Fatalf() is what that'd be for

ldemailly: like the test just above does, changed in #172

ccoVeille: Thanks

PR #168, opened 2024-08-20
Show parsing errors in context of line where they occur

### LINKED ISSUE #15: Better errors, line numbers, dump the ast etc
parser errors in particular lack context

and for other, like using the wrong function name, it's also obscure 

at least should say which is wrong, eg
```
$ typo("abc")
== Parse ==> typo("abc")
== Macro ==> typo("abc")
== Eval  ==> <err: <not a function: ERROR>>
```

### LINKED ISSUE #166: limit number of parsing errors
when parsing errors start specially on a large block (like someone pasting go code into grol bot) should limit errors to first few...

could be a grol setting or I guess handled in discord bot

related to lack of line numbers etc too

ldemailly:
Now:
```
$ if true {
>    len{}
13:04:44.844 [ERR] parser error: expected next token to be `(`, got `{` instead:
   len{}
      ^
$ 1e+999 - 6
13:05:15.404 [ERR] parser error: could not parse "1e+999" as float:
1e+999 - 6
     ^
$ if foo {  % blah
13:07:21.791 [ERR] parser error: no prefix parse function for `%` found:
if foo {  % blah
          ^
```
Fixes #15  (at least for parsing errors, for eval time we may want a stack trace)
Fixes #166  (will follow up in discord bot PR)

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -327,7 +327,7 @@ func evalIndexExpression(left, index object.Object) object.Object {
 		if idx < 0 || idx >= int64(len(str)) {
 			return object.NULL
 		}
-		return object.Integer{Value: int64(str[idx])}
+		return object.Integer{Value: int64(str[idx])} //nolint:gosec // https://github.com/securego/gosec/issues/1185
 	case left.Type() == object.ARRAY && idxOrZero.Type() == object.INTEGER:
 		return evalArrayIndexExpression(left, idxOrZero)
 	case left.Type() == object.MAP:
@@ -725,7 +725,7 @@ func evalIntegerInfixExpression(operator token.Type, left, right object.Object) 
 	case token.LEFTSHIFT:
 		return object.Integer{Value: leftVal << rightVal}
 	case token.RIGHTSHIFT:
-		return object.Integer{Value: int64(uint64(leftVal) >> rightVal)}
+		return object.Integer{Value: int64(uint64(leftVal) >> rightVal)} //nolint:gosec // we want to be able to shift the hight bit.
 	case token.BITAND:
 		return object.Integer{Value: leftVal & rightVal}
 	case token.BITOR:
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -45,11 +45,12 @@ type Parser struct {
 	curToken  *token.Token
 	peekToken *token.Token
 
-	prevNewline bool
-	nextNewline bool
-
-	errors             []string
+	prevNewline        bool
+	nextNewline        bool
 	continuationNeeded bool
+	prevPos            int
+
+	errors []string
 
 	prefixParseFns  map[token.Type]prefixParseFn
 	infixParseFns   map[token.Type]infixParseFn
@@ -155,6 +156,7 @@ func (p *Parser) Errors() []string {
 func (p *Parser) nextToken() {
 	p.prevToken = p.curToken
 	p.curToken = p.peekToken
+	p.prevPos = p.l.Pos()
 	p.peekToken = p.l.NextToken()
 	p.prevNewline = p.nextNewline
 	p.nextNewline = p.l.HadNewline()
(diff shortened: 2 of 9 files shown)


eval/eval.go
```diff
@@ -327,7 +327,7 @@ func evalIndexExpression(left, index object.Object) object.Object {
 		if idx < 0 || idx >= int64(len(str)) {
 			return object.NULL
 		}
-		return object.Integer{Value: int64(str[idx])}
+		return object.Integer{Value: int64(str[idx])} //nolint:gosec // https://github.com/securego/gosec/issues/1185
```
ccoVeille, line 330: A comment explaining the problem would be appreciated. Yes, there is a link, and it's great. but someone reading needs tpneedo open the link. People are lazy. I can tell I'm lazy 😬

parser/parser.go
```diff
@@ -273,14 +275,25 @@ func (p *Parser) expectPeek(t token.Type) bool {
 	return false
 }
 
+// ErrorContext returns the current line and a pointer to the error position.
+// If prev is true, the error position is relative to the previous token instead of current one.
+func (p *Parser) ErrorContext(prev bool) string {
```
ccoVeille, line 280: Two remarks:
- context is connoted in Go. I would have avoided
- it's not really about error. You do use it to say where the error is, but there is nothing about error here. The function can be called even when there is no error

parser/parser.go
```diff
@@ -273,14 +275,25 @@ func (p *Parser) expectPeek(t token.Type) bool {
...
+	line, errPos := p.l.CurrentLine()
+	if prev {
+		errPos -= (p.l.Pos() - p.prevPos)
```
ccoVeille, line 283: This line require a comment as the computing you do is cryptic.

People don't read the function comment.

This line should be documented

parser/parser.go
ccoVeille: Maybe tgis method could call Errors() (it's the right object, right ?) and return a different result when there is no error

parser/parser_test.go
```diff
@@ -364,3 +364,18 @@ func TestIncompleteBlockComment(t *testing.T) {
 		}
 	}
 }
+
+func TestNilToken(t *testing.T) {
```
ccoVeille, line 368: I don't think people will understand what this test is about

token/token.go
```diff
@@ -308,12 +309,16 @@ func LookupIdent(ident string) *Token {
...
-// (even though that's better than string comparaisons).
+// (even though that's better than string comparisons).
 func ByType(t Type) *Token {
 	return tToT[t]
 }
 
 func (t *Token) Literal() string {
+	if t == nil {
+		log.Critf("Nil token .Literal() called")
+		return "NIL_TOKEN"
```
ccoVeille, line 320: Is there a risk if someone write NIL_TOKEN and ask grol to parse it ?

ccoVeille:
I would like to recommend you

https://github.com/meyermarcel/annot

(Hi @meyermarcel 👋)

eval/eval.go
ldemailly: strings[idx] are bytes (unsigned uint8) and the linter thinks that 0-255 value can't fit into a int64 which is completely stupid

parser/parser.go
ldemailly: it's used to construct the errors so it can't use the errors

token/token.go
ldemailly: no it's if you lookup a token in a map and get nil and then call this method (in this case it was @ being looked up and not found in the map and then still used to print error, I fixed that but left the log for the next time a bug like that shows up (though the natural panic NPE might be fine too)

ldemailly:
> I would like to recommend you
> 
> https://github.com/meyermarcel/annot
> 
> (Hi @meyermarcel 👋)

fancy :) I think I'm ok with my lowkey single ^ though for now :)

parser/parser_test.go
ldemailly: changing to this in #171 
```go
func TestInvalidToken(t *testing.T) {
	// This turns into the invalid token which in turn can't be found in prefix map,
	// it used to crash in earlier versions, this is the regression for that crash.
	// it also checks the error message in first line + first column case.
	inp := "@"
```

parser/parser.go
ldemailly: doing in #171  - note that if you don't read function comment... yet read code/non function comment... seems a bit odd / spurious argument

parser/parser_test.go
ccoVeille: OK 👍

eval/eval.go
ldemailly: the link to that bug is enough. and hopefully it get fixed, though it pisses me off golanglint-ci action pin doesn't pin the linters themselves and stuff start breaking mid PRs </rant>

ccoVeille:
> > I would like to recommend you
> > 
> > https://github.com/meyermarcel/annot
> > 
> > (Hi @meyermarcel 👋)
> 
> fancy :) I think I'm ok with my lowkey single ^ though for now :) 

Noted. I also found https://github.com/caarlos0/mdtree

New and interesting

parser/parser.go
ccoVeille: I know 😬

PR #171, opened 2024-08-21
Index range expression, including right relative to the end for both direct access and ranges

ldemailly:
```go
$ "ABCDEF"[2:-1]
"CDE"

$ 12:16
[12,13,14,15]

$ (12:16)[2:] 
[14,15]

```
working for now (and maps not broken) - did extra work getting `slice[42:]` to work

bonus:
```go
$ -42:-37
[-42,-41,-40,-39,-38]
$ [1,2,3,4,5][-1]
5
$ [1,2,3,4,5][-2:-1]
[4]
// ^ may look odd but consistent with
$ [1,2,3,4,5][1]    
2
$ [1,2,3,4,5][0:1]
[1]
```

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -276,11 +276,13 @@ func (p *Parser) expectPeek(t token.Type) bool {
 	return false
 }
 
-// ErrorContext returns the current line and a pointer to the error position.
+// ErrorLine returns the current line and a pointer to the error position.
 // If prev is true, the error position is relative to the previous token instead of current one.
-func (p *Parser) ErrorContext(prev bool) string {
+func (p *Parser) ErrorLine(prev bool) string {
 	line, errPos := p.l.CurrentLine()
 	if prev {
+		// When the error about the previous tokem, adjust the position accordingly.
+		// (note this doesn't work when the previous token in on a different line -- TODO: improve)
 		errPos -= (p.l.Pos() - p.prevPos)
 	}
 	repeat := max(0, errPos-1)
@@ -289,12 +291,12 @@ func (p *Parser) ErrorContext(prev bool) string {
 
 func (p *Parser) peekError(t token.Type) {
 	msg := fmt.Sprintf("expected next token to be `%s`, got `%s` instead:\n%s",
-		token.ByType(t).Literal(), p.peekToken.Literal(), p.ErrorContext(false))
+		token.ByType(t).Literal(), p.peekToken.Literal(), p.ErrorLine(false))
 	p.errors = append(p.errors, msg)
 }
 
 func (p *Parser) noPrefixParseFnError(t *token.Token) {
-	msg := fmt.Sprintf("no prefix parse function for `%s` found:\n%s", t.Literal(), p.ErrorContext(true))
+	msg := fmt.Sprintf("no prefix parse function for `%s` found:\n%s", t.Literal(), p.ErrorLine(true))
 	p.errors = append(p.errors, msg)
 }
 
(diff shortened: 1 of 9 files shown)


parser/parser.go
```diff
@@ -275,11 +276,13 @@ func (p *Parser) expectPeek(t token.Type) bool {
...
 }
 
-// ErrorContext returns the current line and a pointer to the error position.
+// ErrorLine returns the current line and a pointer to the error position.
 // If prev is true, the error position is relative to the previous token instead of current one.
-func (p *Parser) ErrorContext(prev bool) string {
+func (p *Parser) ErrorLine(prev bool) string {
 	line, errPos := p.l.CurrentLine()
 	if prev {
+		// When the error about the previous tokem, adjust the position accordingly.
```
ccoVeille, line 284:
```suggestion
		// When the error about the previous token, adjust the position accordingly.
```

parser/parser.go
```diff
@@ -275,11 +276,13 @@ func (p *Parser) expectPeek(t token.Type) bool {
 	return false
 }
 
-// ErrorContext returns the current line and a pointer to the error position.
+// ErrorLine returns the current line and a pointer to the error position.
 // If prev is true, the error position is relative to the previous token instead of current one.
-func (p *Parser) ErrorContext(prev bool) string {
+func (p *Parser) ErrorLine(prev bool) string {
```
ccoVeille, line 281: This maybe?

```suggestion
func (p *Parser) ErrorLine(previousToken bool) string {
```

ccoVeille, line 281: I like the new name

token/token.go
```diff
@@ -315,10 +314,6 @@ func ByType(t Type) *Token {
 }
 
 func (t *Token) Literal() string {
-	if t == nil {
-		log.Critf("Nil token .Literal() called")
-		return "NIL_TOKEN"
-	}
```
ccoVeille, line 321: Cool

parser/parser.go
ldemailly: fixing in next pr :)


any comments on the harder stuff ... :) ?

parser/parser.go
ldemailly:
```diff
diff --git a/parser/parser.go b/parser/parser.go
index b4e47c9..aa386a5 100644
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -277,11 +277,11 @@ func (p *Parser) expectPeek(t token.Type) bool {
 }
 
 // ErrorLine returns the current line and a pointer to the error position.
-// If prev is true, the error position is relative to the previous token instead of current one.
-func (p *Parser) ErrorLine(prev bool) string {
+// If forPreviousToken is true, the error position is relative to the previous token instead of current one.
+func (p *Parser) ErrorLine(forPreviousToken bool) string {
        line, errPos := p.l.CurrentLine()
-       if prev {
-               // When the error about the previous tokem, adjust the position accordingly.
+       if forPreviousToken {
+               // When the error is about the previous token, adjust the position accordingly.
                // (note this doesn't work when the previous token in on a different line -- TODO: improve)
                errPos -= (p.l.Pos() - p.prevPos)
        }
```

parser/parser.go
ccoVeille: No, I decided it was too complicated, so I would only review the obvious oddities 🤪 🤡

parser/parser.go
ccoVeille: Better indeed

ldemailly: cool it's in #172

PR #172, opened 2024-08-21
Adding => for short lambda form. Bug fixes with incomplete maps.

### LINKED ISSUE #135: (possibly) add `=>` for shorter lambdas
eg

```go
x => x*x
```
would be equivalent to
```go
func (x) {x*x}
```

ldemailly:
- Adding => for short form of lambda ( `x => x*2` == `func (x) {x*2}`)
- Fixed crash with incomplete map.
- Fix bug with incomplete input not reported as error when using grol -c / string eval. 
- Earlier review comment tweaks

Fixes #135


eval/eval.go
```diff
@@ -171,6 +171,11 @@ func (s *State) evalInternal(node any) object.Object { //nolint:funlen,gocyclo /
 			Env:        s.env,
 			Body:       node.Body,
 			Variadic:   node.Variadic,
+			Lambda:     node.IsLambda,
+		}
+		if !fn.Lambda && len(fn.Parameters) == 1 && len(fn.Body.Statements) == 1 && fn.Name == nil {
+			log.LogVf("Normalizing non short lambda form to => lambda")
```
ccoVeille, line 177:
```suggestion
			log.LogVf("Normalizing non-short lambda form to => lambda")
```

lexer/lexer.go
```diff
@@ -68,6 +68,10 @@ func (l *Lexer) NextToken() *token.Token {
 			// so we normalize := like it didn't exist.
 			return token.ConstantTokenChar2(ch, nextChar)
 		}
+		if nextChar == '>' && ch == '=' {
```
ccoVeille, line 71: Maybe you could invert


```suggestion
		if ch == '=' && nextChar == '>' {
```

Or add a comment


```suggestion
		// detecting lamda function ( '=>' )
		if nextChar == '>' && ch == '=' {
```

Or something like that

main_test.txtar
```diff
@@ -204,6 +204,12 @@ stdout '^\[27,28,29,30\]$'
 grol -quiet -c '(23:31)[5-1:23-24]'
 stdout '^\[27,28,29\]$'
 
+# crash on weird half function body
+!grol -quiet -c '(x) {1'
+stderr 'Incomplete input'
+!stderr 'panic'
+!stderr 'runtime error: invalid memor'
```
ccoVeille, line 211: Shouldn't be this?

```suggestion
!stderr 'runtime error: invalid memory'
```

lexer/lexer.go
ldemailly: I added a comment, the test is in the order it is because ch is already likely to be = (or ! ...) so I want the most specific first and no retest unless needed

eval/memo_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"grol.io/grol/eval"
+	"grol.io/grol/object"
+)
+
+func TestBigArrayNotHashable(t *testing.T) {
+	// This test is to make sure that big arrays are no hashable.
```
ldemailly, line 11:
```suggestion
	// This test is to make sure that big arrays are showing as not hashable (so no crash when used later).
```

ldemailly:
I may bite the bullet and make

(x,y,z) => x+y+z

but that's a big change to the parser

token/token.go
```diff
@@ -114,6 +114,7 @@ const (
 	AND
 	LEFTSHIFT
 	RIGHTSHIFT
+	LAMBDA // => lambda short cut: `a,b => a+b`` alias for `func(a,b) {a+b}``
```
ccoVeille, line 117: I don't get the double backtick, maybe  you wanted this

```suggestion
	LAMBDA // => lambda short cut: `a,b => a+b` alias for `func(a,b) {a+b}`
```

ccoVeille (approved):
One minor remark, but seems OK

ldemailly:
> One minor remark, but seems OK

you meant awesome :) 

thanks for the reviews!

object/object.go
```diff
@@ -489,6 +489,7 @@ type Function struct {
 	Body       *ast.Statements
 	Env        *Environment
 	Variadic   bool
+	Lambda     bool // no name, 1 param, 1 statement.
```
ldemailly, line 492:
```suggestion
	Lambda     bool // i.e. has no name.
```

PR #174, opened 2024-08-23
Pin linter version and fix nolintlint/gosec fix

ldemailly:
Same as https://github.com/fortio/workflows/pull/67 (need to figure cross org share actually rather than copy-pasta)

Pending https://github.com/golangci/golangci-lint-action/issues/1091


.github/workflows/gochecks.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         run: curl -fsS -o .golangci.yml https://raw.githubusercontent.com/fortio/workflows/main/golangci.yml
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # pin@v6
+        with:
+          version: 1.60.3
```
ldemailly, line 31:
```suggestion
          version: v1.60.3
```

PR #175, opened 2024-08-23
Unwrap() no panic even with map of unhashable keys and option to force map keys to be strings, adding json_go with optional indent

ldemailly:
we could build tag the encoding/json, it adds 152k
```
├─────────┬──────────────────────────┬──────────┬──────────┬─────────┤
│ PERCENT │ NAME                     │ OLD SIZE │ NEW SIZE │ DIFF    │
├─────────┼──────────────────────────┼──────────┼──────────┼─────────┤
│ +100%   │ encoding/json            │          │ 83 kB    │ +83 kB  │
│ +64.94% │ slices                   │ 19 kB    │ 32 kB    │ +13 kB  │
│ +7.66%  │ reflect                  │ 66 kB    │ 71 kB    │ +5.1 kB │
│ +16.46% │ sync                     │ 22 kB    │ 25 kB    │ +3.6 kB │
[...]
│ +6.44%  │ grol                     │ 2.4 MB   │ 2.5 MB   │ +152 kB │
```

```go
println(json_go({"abc":42, 63:63, "x": {[3]:122, true:false} }, "  "))
```
```json
{
  "63": 63,
  "abc": 42,
  "x": {
    "[3]": 122,
    "true": false
  }
}
```

### DIFF
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -326,7 +326,7 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
 		ArgTypes: []object.Type{object.ANY},
 		Callback: func(_ any, _ string, args []object.Object) object.Object {
 			o := args[0]
-			switch o.Type() { //nolint:exhaustive // that's what default is for.
+			switch o.Type() {
 			case object.INTEGER:
 				return o
 			case object.NIL:
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -138,22 +138,22 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
 	object.AddIdentifier("PI", object.Float{Value: math.Pi})
 	object.AddIdentifier("E", object.Float{Value: math.E}) // using uppercase so "e" isn't taken/shadowed.
 	jsonFn := object.Extension{
-		Name:     "json_go",
+		Name:     "json",
 		MinArgs:  1,
 		MaxArgs:  1,
 		ArgTypes: []object.Type{object.ANY},
-		Callback: object.ShortCallback(jsonSerGo),
+		Callback: object.ShortCallback(jsonSer),
 	}
 	err = object.CreateFunction(jsonFn)
 	if err != nil {
 		return err
 	}
 	jsonFn = object.Extension{
-		Name:     "json",
+		Name:     "json_go",
 		MinArgs:  1,
-		MaxArgs:  1,
-		ArgTypes: []object.Type{object.ANY},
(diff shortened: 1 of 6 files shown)


extensions/extension.go
```diff
@@ -147,6 +148,17 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
 	if err != nil {
 		return err
 	}
+	jsonFn = object.Extension{
+		Name:     "json_go",
```
ccoVeille, line 152: Why is "json_go" identifier?

object/object.go
```diff
@@ -80,6 +82,15 @@ func Hashable(o Object) bool {
 	return false
 }
 
+func UnwrapHashable(o Object) any {
+	switch o.Type() { //nolint:exhaustive // getting a bit tired of this linter, why does it trigger when you have a default?
```
ccoVeille, line 86: I hate this linter, I always disable it

ccoVeille (approved):
LGTM 👍

extensions/extension.go
ldemailly: not sure... I have `json` where the serialization is done by hand (and uses a few extra tags for functions and errors), this one is using go's marshaller (and the new unwrap)

suggestion for a better name?

object/object.go
ldemailly: I'd like to have it only when there is no "default" - also I'd like it to work on type switch but it doesn't

I should disable it or find how to tune it so default makes it shut up

ldemailly: AH!     default-signifies-exhaustive: true
https://golangci-lint.run/usage/linters/#exhaustive

ldemailly: https://github.com/fortio/workflows/pull/68

extensions/extension.go
ccoVeille: I wouldn't use json_go as grol shouldn't mention Go

Can you use a to_json, object, array, to_object, json_object, or from_json ?

I'm unsure the usage of the operator
Wording is a pain

ldemailly: naming is hard. grol being in go (so far) is a bit part of the language (Go REPL...) but maybe more relevant: the extension mechanism does let you map any go function to grol (and bit like Tcl did with C back in the days)

I want to keep json() maybe I can make this one json_pp (pretty print) but the pretty print is just an option (the additional optional string padding argument)

both `json()` and `json_go()` are functions (not operators btw) - there is `unjson()` which is really... eval() atm but that one maybe could be named `from_json()` instead; that doesn't help me name the 2 `json`s though :)

PR #177, opened 2024-08-23
Bug with map returned from lamba single arg, missing outer braces that are needed

ldemailly:
Started as a more complex one in discord bot but reduced to :

```
$ ()=>{{"a":1,"b":2}}
()=>{"a":1,"b":2}
```
which doesn't work when reloaded:
```
parser error: no prefix parse function for `,` found:
()=>{"a":1,"b":2}
          ^
```

- [x] test failing
- [x] fix


ccoVeille (approved):
LGTM 👍

PR #178, opened 2024-08-23
Optional ClientData and Help for extensions, PreInput hook so per interp client data can be set, expose Extensions

ldemailly:
Needed (https://github.com/grol-io/grol-discord-bot/pull/83 which 'tests' this) to make grol discord bot safer for sending complex messages where the channelID is attached to the ClientData and thus can't be modified by grol code

Extensions can be changed/set using now exposed Extensions in the state.

Improves on extensions discoverable help; eg:

```go
$ json_go  
json_go(any, [string]) // optional indent e.g json_go(m, "  ")
$ save
save([string]) // filename (.gr)
```

Through optional Help field

And client data replaces the default eval.State first argument in CallBack

Also adding repl.EvalStringOptions() which has the necessary starting values for EvalStringWithOptions to function.

### DIFF
--- a/eval/eval_api.go
+++ b/eval/eval_api.go
@@ -22,7 +22,7 @@ type State struct {
 	macroState *object.Environment
 	env        *object.Environment
 	cache      Cache
-	extensions map[string]object.Extension
+	Extensions map[string]object.Extension
 	NoLog      bool // turn log() into println() (for EvalString)
 	// Max depth / recursion level - default DefaultMaxDepth,
 	// note that a simple function consumes at least 2 levels and typically at least 3 or 4.
@@ -38,7 +38,7 @@ func NewState() *State {
 		Out:        os.Stdout,
 		LogOut:     os.Stdout,
 		cache:      NewCache(),
-		extensions: object.ExtraFunctions(),
+		Extensions: object.ExtraFunctions(),
 		macroState: object.NewMacroEnvironment(),
 		MaxDepth:   DefaultMaxDepth,
 		depth:      0,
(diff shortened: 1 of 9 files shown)


repl/repl.go
```diff
@@ -223,7 +232,9 @@ func Interactive(options Options) int {
 	term.NewHistory(options.MaxHistory)
 	_ = term.SetHistoryFile(options.HistoryFile)
 	_ = AutoLoad(s, options) // errors already logged
-	// Regular expression for "!nn" to run history command nn.
+	if options.PreInput != nil {
```
ccoVeille, line 235: No test for this?

eval/eval_api.go
```diff
@@ -22,7 +22,7 @@ type State struct {
 	macroState *object.Environment
 	env        *object.Environment
 	cache      Cache
-	extensions map[string]object.Extension
+	Extensions map[string]object.Extension
```
ccoVeille, line 25: I don't see why you are changing the visibility.

repl/repl.go
ldemailly: the test (use) is in the bot but I can probably should add one here

eval/eval_api.go
ldemailly: because that's the most direct way to allow additions in the preinput hook, the test (or the other pr meanwhile) will make that more clear

repl/repl.go
ldemailly: added, ptal

eval/eval_api.go
ldemailly: added test showing how it's used/why

ldemailly:
I have more stuff I want to change/fix but will merge this one first and make smaller additional PRs

ccoVeille (approved):

PR #182, opened 2024-08-26
Assignment to map keys and array indexes

### LINKED ISSUE #16: assignment to maps keys and arrays indexes
```go
a=[1,2,3]
a[0]=42
```
`<err: <assignment to non identifier: (a[0])>>`
shud instead 'work' and mutate `a`

ldemailly:
- [x] Test
- [x] Fix/impl

Fixes #16

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -19,16 +19,68 @@ var unquoteToken = token.ByType(token.UNQUOTE)
 
 func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) object.Object {
 	// let free assignments.
-	id, ok := node.Left.(*ast.Identifier)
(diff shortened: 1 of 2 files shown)


ccoVeille (commented):
You are only testing what is expected to work and not the possible error such as a[10+2]

ldemailly:
> You are only testing what is expected to work and not the possible error such as a[10+2]

a[10+2] is fine (it's a[12]) but if you mean out of bounds check, sure, though I was wondering if I should make them work at least in term of growing the arrays

ccoVeille:
Yes, I'm thinking about out of bound, as 10 is not defined in you example of a with a few items.

PR #184, opened 2024-08-27
Better short circuit upon errors, info.all_ids is array of array of ids instead of key value pairs. Output the line number as part of parsing errors; Stack trace on runtime errors

ldemailly:
- Output the line number as part of parsing errors
- fix autocomplete for info, add history to autocomplete
- remove values from info.all_ids, make error abort sooner in expressions, better error message for invalid operations
- stack trace on errors

```go
$ func level1(){1+"x"}; func level2(){level1()}; func level3(){level2()}; level3()
<err: no PLUS on left=1 right="x", stack below:>
func level1(){1+"x"}
func level2(){level1()}
func level3(){level2()}
$ level1()
<err: no PLUS on left=1 right="x" in func level1(){1+"x"}>
$ 1+"a"
<err: no PLUS on left=1 right="a">
```

### DIFF
--- a/eval/stack.go
+++ b/eval/stack.go
@@ -1,0 +1,35 @@
+package eval
+
+import (
+	"fmt"
+
+	"fortio.org/log"
+	"grol.io/grol/object"
+)
+
+// Returns a stack array of the current stack.
+func (s *State) Stack() []string {
+	stack := make([]string, 0, s.depth) // will be -1 because no name at toplevel but... it's fine.
+	e := s.env
+	for {
+		if e == nil {
+			break
+		}
+		n := e.Name()
+		if n != "" {
+			stack = append(stack, e.Name())
+		}
+		e = e.Parent()
+	}
+	log.Debugf("Stack() len %d, depth %d returning %v", len(stack), s.depth, stack)
+	return stack
+}
+
+// Creates a new error object with the given message and stack.
+func (s *State) Error(msg string) object.Error {
+	return object.Error{Value: msg, Stack: s.Stack()}
+}
+
+func (s *State) Errorf(format string, args ...interface{}) object.Error {
+	return s.Error(fmt.Sprintf(format, args...))
+}
--- a/eval/stack.go
+++ b/eval/stack.go
@@ -19,7 +19,7 @@ func (s *State) Stack() []string {
 		if n != "" {
 			stack = append(stack, e.Name())
 		}
-		e = e.Parent()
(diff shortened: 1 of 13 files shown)


eval/stack.go
```diff
@@ -0,0 +1,35 @@
...
+	return stack
+}
+
+// Creates a new error object with the given message and stack.
+func (s *State) Error(msg string) object.Error {
+	return object.Error{Value: msg, Stack: s.Stack()}
+}
+
+func (s *State) Errorf(format string, args ...interface{}) object.Error {
+	return s.Error(fmt.Sprintf(format, args...))
```
ccoVeille, line 34: I would have used this

```suggestion
	return object.Error{Value: (fmt.Sprintf(format, args...), Stack: s.Stack()}
```

To do not add one more call to the stack. Unless I missed something

extensions/extension.go
```diff
@@ -398,10 +399,11 @@ func jsonSerGo(args []object.Object) object.Object {
...
 
 func evalFunc(env any, name string, args []object.Object) object.Object {
-	s := args[0].(object.String).Value
-	res, err := eval.EvalString(env, s, name == "unjson" /* empty env */)
+	str := args[0].(object.String).Value
+	s := env.(*eval.State)
+	res, err := eval.EvalString(s, str, name == "unjson" /* empty env */)
 	if err != nil {
-		return object.Error{Value: err.Error()}
+		return s.Error(err.Error())
```
ccoVeille, line 406: This pattern is repeated.

You could use `any` the in s.Error, instead of string, the use fmt.Sprint in s.Error to convert the any into a string, allowing to pass an error

ccoVeille (approved):

extensions/extension.go
ldemailly: I used a big regexp search and replace, then noticed the Sprintf variants and replaced that with Errorf but yeah I might as well take error in too

eval/stack.go
ldemailly: well this should be inlined by the compiler as it sees fit, also with the logging added it would make the 2 diverge

extensions/extension.go
ldemailly: `any` is a bit ugly though as I can then accidentally pass something wrong

ldemailly: will add
```go
func (s *State) ErrToError(err error) object.Error {
	return s.Error(err.Error())
}
```
instead

PR #185, opened 2024-08-27
Bug fix: errors aren't hashable anymore, avoid panic; adding missing null for json to object; reduced stack limit; improve Stack()'s implementation

ldemailly:


eval/stack.go
```diff
@@ -9,24 +9,25 @@ import (
...
 	}
 	log.Debugf("Stack() len %d, depth %d returning %v", len(stack), s.depth, stack)
 	return stack
 }
 
 // Creates a new error object with the given message and stack.
 func (s *State) Error(msg string) object.Error {
+	if log.LogDebug() {
+		log.LogVf("Error %q called", msg)
+	}
```
ccoVeille, line 30: Don't you have a log.DebugVf ?

Or the debug mode is conditioned by the code?

I would expect to have a method in log to check .logdebug and do not log if needed

eval/stack.go
```diff
@@ -9,24 +9,25 @@ import (
...
-	e := s.env
-	for {
-		if e == nil {
-			break
-		}
+	if s.depth <= 1 {
+		return nil
+	}
+	stack := make([]string, 0, s.depth-1)
+	for e := s.env; e != nil; e = e.StackParent() {
```
ccoVeille, line 16: Can't stackparent returna nil?

Then the very next call to e.Name will panic

eval/stack.go
ldemailly: I was trying to make the compiler or runtime smarter and skip the log unless log level is debug (and yes I use verbose after which is above debug, mostly because of the blue color it gets on console)

so yes that log.LogDebug() returns a boolean

https://github.com/fortio/log/blob/v1.16.0/logger.go#L566-L568

I should use log.LogVerbose() though

eval/stack.go
ldemailly: > Can't stackparent return a nil?
yes it does and no it won't because that's how we stop the loop :)

ccoVeille: That's right

for (init, check, iterate)

Iterate is called before check right?

ldemailly: iterate is called after each iteration yes, sleep deprivation is doing a number on you :)

PR #186, opened 2024-08-27
Bug with some lambdas and json_go

ldemailly:
- [x] test
- [x] fix
- [x] review comments from earlier PRs

### DIFF
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -347,7 +347,7 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
 			case object.STRING:
 				i, serr := strconv.ParseInt(o.(object.String).Value, 0, 64)
 				if serr != nil {
-					return s.ErrToError(serr)
+					return s.ConvertError(serr)
 				}
 				return object.Integer{Value: i}
 			default:
@@ -381,7 +381,7 @@ func jsonSer(env any, _ string, args []object.Object) object.Object {
 	w := strings.Builder{}
 	err := args[0].JSON(&w)
 	if err != nil {
-		return s.ErrToError(err)
+		return s.ConvertError(err)
 	}
 	return object.String{Value: w.String()}
 }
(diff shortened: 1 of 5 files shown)


main_test.txtar
```diff
@@ -267,6 +267,12 @@ stderr '^func level3\(\){level2\(\)}$'
 stderr 'Total 1 error'
 stderr '^<err: no PLUS on left=1 right="x" in func level1\(\){1\+"x"}>$'
 
+# json with some functions
+grol -quiet -c 'f=()=>{{"k":"v"}}; println(json_go(f))'
+!stderr 'json: unsupported type'
+!stdout 'json: unsupported type'
+stdout '^"\(\)=>{{\\"k\\":\\"v\\"}}"$'
```
ldemailly, line 274: regexp escaping is getting a bit tough/annoying, the output is
```
"()=>{{\"k\":\"v\"}}"
```

extensions/extension.go
```diff
@@ -458,21 +463,21 @@ func loadFunc(env any, _ string, args []object.Object) object.Object {
...
 	all, err := io.ReadAll(f)
 	if err != nil {
-		return s.Error(err.Error())
+		return s.ErrToError(err)
 	}
 	// Eval the content.
 	res, err := eval.EvalString(env, string(all), false)
 	if err != nil {
-		return s.Error(err.Error())
+		return s.ErrToError(err)
```
ldemailly, line 480: @ccoVeille I bet you probably don't like that name, suggestion for a better one?

ccoVeille: I do dislike it.

ReportError ?

ReturnError?

ldemailly: overlord suggested a few, I picked ConvertError() but I could be convinced to use something else

not ReportError as it's not what it does... it makes an error object (which cause evaluation to abort)
```go
return s.ReturnError()
```
looks a bit stuttering :)

ldemailly: maybe I can make the other one ErrorString() and use Error for this one

ldemailly: I think this is actually decent now:
```go
/ NewError creates a new error object from a plain string.
// NewError will attach the stack trace to the NewError object.
func (s *State) NewError(msg string) object.Error
// Errorf formats an object.Error with the given format and args.
func (s *State) Errorf(format string, args ...interface{}) object.Error 
// Error converts from a go error to an object.Error.
func (s *State) Error(err error) object.Error 
```

ccoVeille (approved):

PR #191, opened 2024-08-29
Added `type()`, fix 2 bugs, added triangle example

### LINKED ISSUE #188: parsing issue/weirdness with literal arrays in if
```
if true{log("x")[i,j]}
14:53:22.838 [ERR] parser error: 1: expected next token to be `]`, got `,` instead:
if true{log("x")[i,j]}
                  ^
14:53:22.838 [ERR] parser error: 1: no prefix parse function for `,` found:
if true{log("x")[i,j]}
                  ^
14:53:22.838 [ERR] parser error: 1: no prefix parse function for `]` found:
if true{log("x")[i,j]}
                    ^
```

```
()=>if true{log("x");[i,j]} //parses fine with ; but then generates the above which doesn't parse back
()=>if true{log("x")[i,j]}
```

probably because of index operator

### LINKED ISSUE #190: panic: strings: negative Repeat count
!grol "x"*-1

### LINKED ISSUE #192: adding `type()` function
It would be neat to know the `type` of an argument. Dealing with arrays in runtime is particularly demanding case for such thing.

ldemailly:
- [x] triangle example
- #190 
- #188 (which is really... issue with literal arrays in general)
- #192 

Fixes #190 
Fixes #188
Fixes #192


ast/ast.go
```diff
@@ -187,8 +187,8 @@ func prettyPrintCompact(ps *PrintState, s Node, i int) bool {
 		return true
 	}
 	_, prevIsExpr := ps.prev.(*InfixExpression)
-	// _, curIsID := s.(*Identifier)
-	if prevIsExpr && ps.last != "}" && ps.last != "]" {
+	_, curIsArray := s.(*ArrayLiteral)
+	if (prevIsExpr && ps.last != "}" && ps.last != "]") || curIsArray {
```
ccoVeille, line 191:
```suggestion
	if curIsArray || (prevIsExpr && ps.last != "}" && ps.last != "]") {
```
What about this?

eval/eval.go
```diff
@@ -785,6 +785,9 @@ func evalStringInfixExpression(operator token.Type, left, right object.Object) o
 	case operator == token.ASTERISK && right.Type() == object.INTEGER:
 		rightVal := right.(object.Integer).Value
 		n := len(leftVal) * int(rightVal)
+		if rightVal < 0 {
+			return s.NewError("right operand of * on strings must be a positive integer")
```
ccoVeille, line 789: Is there a need for the zero?

parser/parser.go
```diff
@@ -314,6 +314,11 @@ func (p *Parser) parseExpression(precedence ast.Priority) ast.Node {
 			log.LogVf("parseExpression: call expression with whitespace")
 			return leftExp
 		}
+		// same for a [n] that's the [n] literal array.
+		if t == token.LBRACKET && p.l.HadWhitespace() {
+			log.LogVf("parseExpression: index expression with whitespace")
```
ccoVeille, line 319: Logging the content might be interesting no?

ast/ast.go
ldemailly: sure it's probably easier to read

eval/eval.go
ldemailly: yes *0 is useful/valid and gets empty string (or empty array or empty map)

PR #194, opened 2024-08-29
`for expr {body}` loops

### LINKED ISSUE #18: for loop
we now have for.gr (#58) but maybe an actual iterative version instead of recursive would be nice

ldemailly:
Fixes #18 

some of the syntax is a bit controversial

```go
$ for 10 {...will do something 10 times...}
$ for i<3 {... normal condition based execution...}
$ for i = -3:5 {... range starts at -3 end at 5-1...}
$ for kv = {"a":1, PI:2} {println("key", kv.key, "value", kv.value)}  
key 3.141592653589793 value 2
key a value 1
$ for r = "abcd" {println(r)}
a
b
c
d
```

Bug fixes
- cpu profiling was broken because of defer
- post panic reset needed to get back to root env (in addition to depth reset)


eval/eval.go
```diff
@@ -672,6 +673,124 @@ func (s *State) evalIfExpression(ie *ast.IfExpression) object.Object {
...
+func (s *State) evalForSpecialForms(fe *ast.ForExpression) (object.Object, bool) {
+	ie, ok := fe.Condition.(*ast.InfixExpression)
+	if !ok {
+		return object.NULL, false
+	}
+	if ie.Token.Type() != token.ASSIGN {
+		return object.NULL, false
+	}
+	if ie.Left.Value().Type() != token.IDENT {
+		return s.Errorf("for var = ... not a var %s", ie.Left.Value().DebugString()), true
```
ccoVeille, line 709: Do you have a kind of language words you forbid as syntax variable.

Something like for, and, or, 

It would avoid strange syntax such as

for for = [0:5]

eval/eval.go
```diff
@@ -672,6 +673,124 @@ func (s *State) evalIfExpression(ie *ast.IfExpression) object.Object {
...
+			log.LogVf("for %s is object.FALSE, done", fe.Condition.Value().DebugString())
+			return lastEval
+		default:
+			switch condition.Type() {
+			case object.ERROR:
+				return condition
+			case object.INTEGER:
+				return s.evalForInteger(fe, nil, condition.(object.Integer), "")
+			default:
+				return s.NewError("for condition is not a boolean nor integer nor assignment: " + condition.Inspect())
```
ccoVeille, line 784:
```suggestion
				return s.NewError("for condition is not a boolean, integer, nor assignment: " + condition.Inspect())
```

The Oxford list version would be this,no?

eval/eval.go
ldemailly: well `for for`does fail because `for` is a keyword

reserved are
```
$ info.keywords
["else","false","for","func","if","return","true"]
$ info.builtins
["error","first","len","log","macro","print","println","quote","rest","unquote"]
```

eval/eval.go
ldemailly: Screenshot 2024-08-29 at 12 11 17 PM

ldemailly: so... I'll leave it, it also said when asked "is one a oxford list?"
"
Yes, the first option is an example of a list that uses an Oxford comma:

**"For condition is not a boolean, integer, nor assignment."**

The Oxford comma is the comma placed before the conjunction ("nor" in this case) in a list of three or more items. However, in your first option, the comma before "nor" isn't actually functioning as a typical Oxford comma because it's followed by "nor," not "and" or "or," which makes it somewhat unconventional.

The second option:

**"For condition is not a boolean nor integer nor assignment."**

does not use the Oxford comma and avoids potential confusion, making it clearer and more direct.
"

PR #196, opened 2024-08-29
Add rand() and fix memoization bug. Change that extensions can't be changed

ldemailly:
later is a fairly breaking change but otherwise id look ups are cache misses

### DIFF
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -150,8 +150,8 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
 				return object.Float{Value: rand.Float64()} //nolint:gosec // no need for crypto/rand here.
 			}
 			n := args[0].(object.Integer).Value
-			if n < 0 {
-				return object.Error{Value: "negative argument to rand()"}
+			if n <= 0 {
+				return object.Error{Value: "argument to rand() if given must be > 0, >=2 for something useful"}
 			}
 			return object.Integer{Value: rand.Int64N(n)} //nolint:gosec // no need for crypto/rand here.
 		},
--- a/extensions/extension.go
+++ b/extensions/extension.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand/v2"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


ldemailly:
going to merge this and make a new PR for break/continue

extensions/extension.go
```diff
@@ -137,6 +138,28 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
...
+		MaxArgs:  1,
+		ArgTypes: []object.Type{object.INTEGER},
+		Callback: func(env any, _ string, args []object.Object) object.Object {
+			eval.TriggerNoCache(env)
+			if len(args) == 0 {
+				return object.Float{Value: rand.Float64()} //nolint:gosec // no need for crypto/rand here.
+			}
+			n := args[0].(object.Integer).Value
+			if n <= 0 {
+				return object.Error{Value: "argument to rand() if given must be > 0, >=2 for something useful"}
```
ccoVeille, line 154: I don't get it. It's because it's a variadic right?

The error message is unclear

I mean if n<=0,means you expect arguments,

But then the message talk about >=2

Why not checking >=2 then?

extensions/extension.go
```diff
@@ -137,6 +138,28 @@ func initInternal(c *Config) error { //nolint:funlen,gocognit,gocyclo,maintidx /
...
+			}
+			return object.Integer{Value: rand.Int64N(n)} //nolint:gosec // no need for crypto/rand here.
```
ccoVeille, line 156: math/rand/v2 might help

extensions/extension.go
ldemailly: it's not technically an error to do rand(1) but it will always produce 0, yet maybe some code needs a random based on some other value that can change so, like the underlying go function it doesn't fail but if someone does make the mistake of calling with 0 they may read the message and realize they should probably have 2 as the minimum, could be a warning but there isn't a grol side warning

extensions/extension.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand/v2"
```
ldemailly, line 11: ^ @ccoVeille

PR #197, opened 2024-08-29
Adding `break` and `continue` for `for` loops.

### LINKED ISSUE #195: add `break` and `continue` for for

ldemailly:
fun one - cc @MikalaiLappo as an example of adding things (though for operators look maybe at BITXOR for instance, it's simpler)

fixes #195


ldemailly:
could use more tests but did a lot of manual tests

ccoVeille (approved):

PR #206, opened 2024-09-02
Reference object type: Allow mutations in upper scope, add back `:=` for forcing local scope variable creation; cache bug fix

### LINKED ISSUE #207: Allow mutations in upper scope, add back `:=` for forcing local scope variable creation
Until now every capture of state was a copy, it was impossible to modify upper state without collaboration from the caller in term of assigning a result

We introduce capture by reference instead of by value by default now. And `a:=a` to make a local copy (or only use the function arguments which are always local)

```go
a = 2
()=>{a=3}()
a // now 3
```

ldemailly:
Fixes #207 

Until now every capture of state was a copy, it was impossible to modify upper state without collaboration from the caller in term of assigning a result

We introduce capture by reference instead of by value by default now. And `a:=a` to make a local copy (or only use the function arguments which are always local)

```go
a = 2
()=>{a=3}()
a // now 3
```


object/state.go
```diff
@@ -175,17 +175,43 @@ func (e *Environment) SaveGlobals(to io.Writer, maxValueLen int) (int, error) {
...
+		if ok {
+			for obj.Type() == REFERENCE { // for is in theory not needed as we deref new refs.
+				obj = obj.(Reference).Value()
+			}
+			ref := Reference{Name: name, RefEnv: e.outer}
+			orig.store[name] = ref
+			orig.getMiss++ // creating a ref is a miss.
+			return &ref, true
+		}
+		e = e.outer
```
ccoVeille, line 191:
```suggestion
		obj, ok := e.outer.store[name]
		if !ok {
			e = e.outer
			continue
		}
		for obj.Type() == REFERENCE { // for is in theory not needed as we deref new refs.
			obj = obj.(Reference).Value()
		}
		ref := Reference{Name: name, RefEnv: e.outer}
		orig.store[name] = ref
		orig.getMiss++ // creating a ref is a miss.
		return &ref, true
		
```

ldemailly: hmm I just explained why the code shouldn't work :)

ldemailly:
diff is now shorter, no more a million mechanical peppering of `object.Value(x)` in front of `x.(object.Type)`

instead we try to decide to use Eval() when we need to unwrap the refererences (which is a few more place and also somewhat mechanical, see todo)

PR #211, opened 2024-09-04
Add begining image function and demo and performance optimizations (10x improvements related to logging level skips)

ldemailly:
./grol examples/image.gr

Produces

Time elapsed:  0.6636569499969482  seconds
Saved image to grol.png

grol

### DIFF
--- a/extensions/images.go
+++ b/extensions/images.go
@@ -1,0 +1,129 @@
+package extensions
+
+import (
+	"image"
+	"image/color"
+	"image/draw"
+	"image/png"
+	"os"
+
+	"grol.io/grol/object"
+)
+
+type ImageMap map[object.Object]*image.RGBA
+
+// make this configurable and use the slice check as well as some sort of LRU.
+const MaxImageSize = 1024
+
+func int2RGBAColor(o object.Object) (uint8, *object.Error) {
+	if o.Type() != object.INTEGER {
+		return 0, &object.Error{Value: "color component not an integer:" + o.Inspect()}
+	}
+	i := o.(object.Integer).Value
+	if i < 0 || i > 255 {
+		return 0, &object.Error{Value: "color component out of range:" + o.Inspect()}
+	}
+	return uint8(i), nil //nolint:gosec // gosec not smart enough to see the range check just above
+}
+
+func arrayToRBGAColor(arr []object.Object) (color.RGBA, *object.Error) {
+	rgba := color.RGBA{}
+	if len(arr) < 3 || len(arr) > 4 {
+		return rgba, &object.Error{Value: "color array must have 3 or 4 elements"}
+	}
+	var oerr *object.Error
+	rgba.R, oerr = int2RGBAColor(arr[0])
+	if oerr != nil {
+		return rgba, oerr
+	}
+	rgba.G, oerr = int2RGBAColor(arr[1])
+	if oerr != nil {
+		return rgba, oerr
+	}
+	rgba.B, oerr = int2RGBAColor(arr[2])
+	if oerr != nil {
+		return rgba, oerr
+	}
+	if len(arr) == 4 {
+		rgba.A, oerr = int2RGBAColor(arr[3])
+		if oerr != nil {
+			return rgba, oerr
+		}
+	} else {
+		rgba.A = 255
+	}
+	return rgba, nil
+}
+
+func createImageFunctions() {
+	cdata := make(ImageMap)
+	imgFn := object.Extension{
+		Name:       "image",
+		MinArgs:    3,
+		MaxArgs:    3,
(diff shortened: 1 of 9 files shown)


extensions/images.go
```diff
@@ -0,0 +1,129 @@
...
+		Help:       "create a new image of the name and size",
+		ArgTypes:   []object.Type{object.STRING, object.INTEGER, object.INTEGER},
+		ClientData: cdata,
+		Callback: func(cdata any, _ string, args []object.Object) object.Object {
+			images := cdata.(ImageMap)
+			x := int(args[1].(object.Integer).Value)
+			y := int(args[2].(object.Integer).Value)
+			if x > MaxImageSize || y > MaxImageSize {
+				return object.Error{Value: "image size too large"}
+			}
```
ccoVeille, line 73: I had to find this to understand MaxImageSize was in pixel and not bytes

Maybe a name with pixel in it

MaxPixel or MaxImagePixel

extensions/images.go
```diff
@@ -0,0 +1,129 @@
...
+			img := image.NewRGBA(image.Rect(0, 0, x, y))
+			transparent := color.RGBA{0, 0, 0, 0}
+			draw.Draw(img, img.Bounds(), &image.Uniform{transparent}, image.Point{}, draw.Src)
+			images[args[0]] = img
+			return args[0]
+		},
+	}
+	MustCreate(imgFn)
+	imgFn.Name = "image_set"
+	imgFn.Help = "set a pixel in the named image, color is an array of 3 or 4 elements 0-255"
```
ccoVeille, line 86: You should mention RGB and RGBA people might not get what are these 3-4 number

Also I think most people are used to define color in hexa, not in decimal

extensions/images.go
```diff
@@ -0,0 +1,129 @@
...
+	imgFn.MinArgs = 4
+	imgFn.MaxArgs = 4
+	imgFn.ArgTypes = []object.Type{object.STRING, object.INTEGER, object.INTEGER, object.ARRAY}
+	imgFn.Callback = func(cdata any, _ string, args []object.Object) object.Object {
+		images := cdata.(ImageMap)
+		x := int(args[1].(object.Integer).Value)
+		y := int(args[2].(object.Integer).Value)
+		img, ok := images[args[0]]
```
ccoVeille, line 94: Please define a variable for args[0] and args[3]

It's unclear what they are

extensions/images.go
ldemailly: so 0x00-0xFF instead? not sure that's clearer, I was thinking of passing as a single int but I think it's cleaner as triplet - will mention alpha channel

extensions/images.go
```diff
@@ -0,0 +1,129 @@
...
+			transparent := color.RGBA{0, 0, 0, 0}
+			draw.Draw(img, img.Bounds(), &image.Uniform{transparent}, image.Point{}, draw.Src)
+			images[args[0]] = img
+			return args[0]
+		},
+	}
+	MustCreate(imgFn)
+	imgFn.Name = "image_set"
+	imgFn.Help = "set a pixel in the named image, color is an array of 3 or 4 elements 0-255"
+	imgFn.MinArgs = 4
```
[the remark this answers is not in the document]
ldemailly: color isn't optional, it's in the array of color that it can be 3 or 4

extensions/images.go
ccoVeille: you might be right rgba seems to commonly be used with https://www.w3schools.com/css/css3_colors.asp

https://rgbacolorpicker.com/rgba-to-hex

maybe you could keep the current format and accept one param when it starts with #

extensions/images.go
ldemailly: changed

extensions/images.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		outputFile, err := os.Create("grol.png")
+		if err != nil {
+			return object.Error{Value: "error opening image file: " + err.Error()}
+		}
+		err = png.Encode(outputFile, img)
+		if err != nil {
+			return object.Error{Value: "error encoding image: " + err.Error()}
+		}
+		outputFile.Close()
```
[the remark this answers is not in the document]
ldemailly: done ty (even though I don't like defer it's appropriate here (even though I doubt png.Encode ever errors out or if it does that closing the file does anything but should avoid leaks

extensions/images.go
ldemailly: added comment about args[0] being always the image ref/name

args[3] is use directly in a line that has "color" and "array" :)

extensions/images.go
ldemailly: done (adding the other color modes)

ccoVeille (approved):

PR #217, opened 2024-09-06
Added base64, image.png, wasm data: to image, (limited) support for namespaces (virtual map access). moved the image functions under image.*

### LINKED ISSUE #204: wasm (grol.io) bug: pressing run quickly gets into missed time out event loop
image

tiny go version didn't use to do that (run button was disabled until it was done...)

### LINKED ISSUE #212: add image demo / capability to wasm version

ldemailly:
- #212 

Fixes #212 
Fixes #204

### DIFF
--- a/object/interp.go
+++ b/object/interp.go
@@ -2,6 +2,7 @@ package object
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"grol.io/grol/lexer"
@@ -51,6 +52,9 @@ func CreateFunction(cmd Extension) error {
 	name := cmd.Name
 	if len(dotSplit) == 2 {
 		ns, name = dotSplit[0], dotSplit[1]
+		if !ValidIdentifier(ns) {
+			return fmt.Errorf("namespace %q not alphanumeric", ns)
+		}
 	}
 	if !ValidIdentifier(name) {
 		return errors.New(name + ": not alphanumeric")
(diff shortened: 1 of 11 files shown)


object/interp.go
```diff
@@ -26,6 +44,19 @@ func CreateFunction(cmd Extension) error {
 	if cmd.Name == "" {
 		return errors.New("empty command name")
 	}
+	// Only support 1 level of namespace for now.
+	dotSplit := strings.SplitN(cmd.Name, ".", 2)
```
ccoVeille, line 48: This is the kind of thing I'm a bit afraid of

I would expect a method to do the magic of
- encode and do [this](https://github.com/grol-io/grol/pull/217/files#diff-f80b989da1d1c5f67f5ac9041afd2d10aa0810075359ff3109a92f571ef82762R299) 
- and one to split

This way the code is easy to use and if something changes later, it avoids issues

ldemailly: it's just a split by first/only 1 allowed `.` though (and it's covered by the e2e tests) the code you linked in particular needs to stay minimal/inline because of the cost and where it is in code paths

ccoVeille (approved):

object/interp.go
```diff
@@ -2,22 +2,40 @@ package object
...
 
+func ValidIdentifier(name string) bool {
+	if name == "" {
+		return false
+	}
+	for _, b := range []byte(name) {
+		if !lexer.IsAlphaNum(b) {
+			return false
+		}
+	}
```
ccoVeille, line 35: Usually it's better to loop on runes from a string

Unicode characters such as accent, and emoji are badly checked otherwise

Then Unicode package can be used to check the type

ldemailly: it's on purpose (for now at least) only bytes (ascii) is allowed outside of strings

unicode does work fine inside strings (as demoed in many examples)

repl/repl.go
```diff
@@ -380,6 +380,7 @@ func EvalOne(ctx context.Context, s *eval.State, what string, out io.Writer, opt
 		}()
 	}
 	s.SetContext(ctx, options.MaxDuration)
+	defer s.Cancel()
```
ldemailly, line 383: This is the fix for the wasm errors!

PR #219, opened 2024-09-08
Adding vector graphics; add `atan2` as well. Fixed function caching.

ldemailly:
using golang.org/x/image/vector

adding

image.move_to
image.line_to
image.close_path
image.draw

sample generates

grol

difficult part is/was what to do with alpha when adding


extensions/images.go
```diff
@@ -164,7 +171,7 @@ func createImageFunctions() { //nolint:funlen // this is a group of related func
 		Name:       "image.new",
 		MinArgs:    3,
 		MaxArgs:    3,
-		Help:       "create a new RGBA image of the name and size, image starts entirely transparent",
+		Help:       "create a new NRGBA image of the name and size, image starts entirely transparent",
```
ccoVeille, line 174: no one reading this will understand NRGBA :grin:

ldemailly: it's a search and replace, but the distinction is important, it means non pre multiplied ie you can have values of RGB higher than A while in RGBA you can't

ccoVeille: exactly, my point is the fact someone calling the help should get a better explanation than just an obscure acronym :grimacing:

ldemailly: fixed!

ldemailly:
With the line with thickness demo:

grol

ldemailly:
examples/disc.gr:

grol

ldemailly:
circle: thickness 40

grol

thickness 1

grol

ldemailly:
Random bezier points (discord.go version)

image

Issue #223, opened 2024-09-13
macro bug, seems to work once and then get stuck


ldemailly:
```go

unless(5 > 11, println("macro test 1: lower ok"), println("BUG: not lower"))

unless(10 > 5, println("BUG: not greater"), println("macro test 2: greater")) // should output "macro test: greater" and not again "lower ok"
```

actually currently prints
```
macro test: lower ok
macro test: lower ok
```

also unable to get 
```go
NoErr = macro(expr,a,b) {
    quote(
		r = catch(unquote(expr))
	    if r.err {
			error("unexpected error:", r.value)
			unquote(a)
		} else {
			unquote(b)
		}
	)
}
```
to work... complains about expr being unknown...

ldemailly:
it's basically #20

ldemailly:
fixed in #225

Issue #68, opened 2024-07-27
fix macro handling for maps, if the key is changed


ldemailly:
see todo in modify

also #20

ldemailly:
fixed in #225

PR #225, opened 2024-09-13
Copy when ast.Modify() so expanding macros expand each macro use as the right one

### LINKED ISSUE #20: ExpandMacros mutates the original program ~~(maybe it's fine)~~ it's not fine :)
Modify should probably clone or have the option to do so

tbd - then maybe just not return a result (outside of errors I guess)

ldemailly:
Fixes:
- #20 
- #223 
- #68

Was a pretty bad bug and #20 was a smoking gun, so where the comments I then wrote in the code
(basically that macros would only work properly "once")

### DIFF
--- a/ast/modify.go
+++ b/ast/modify.go
@@ -3,7 +3,7 @@ package ast
 import "fmt"
 
 // Note, this is somewhat similar to eval.go's eval... both are "apply"ing.
-func Modify(node Node, f func(Node) Node) Node { //nolint:funlen // yeah lots of types.
+func Modify(node Node, f func(Node) Node) Node { //nolint:funlen,gocyclo // yeah lots of types.
 	// TODO: add err checks for _s.
 	switch node := node.(type) {
 	case *Statements:
--- a/ast/modify.go
+++ b/ast/modify.go
@@ -1,8 +1,6 @@
 package ast
 
-import (
-	"fortio.org/log"
-)
+import "fmt"
 
 // Note, this is somewhat similar to eval.go's eval... both are "apply"ing.
 func Modify(node Node, f func(Node) Node) Node { //nolint:funlen // yeah lots of types.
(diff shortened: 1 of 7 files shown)


ast/modify_test.go
```diff
@@ -119,3 +113,29 @@ func TestModify(t *testing.T) {
 		}
 	}
 }
+
+func TestModifyMap(t *testing.T) {
+	// need to test map separately because deepequal can't compare a map
+	// with keys of the same underlying value (2:2, 2:2)
+	modified := Modify(
+		&MapLiteral{
```
ccoVeille, line 121: DeepEquals cannot, but [maps.Equal](https://pkg.go.dev/maps#Equal) can, no?

ast/modify.go
```diff
@@ -1,55 +1,121 @@
 package ast
 
-import "fortio.org/log"
+import "fmt"
 
 // Note, this is somewhat similar to eval.go's eval... both are "apply"ing.
-func Modify(node Node, f func(Node) Node) Node {
+func Modify(node Node, f func(Node) Node) Node { //nolint:funlen,gocyclo // yeah lots of types.
 	// TODO: add err checks for _s.
```
ccoVeille, line 7: What about a map[string]func(node)

Where you use fmt.Sprintf("%T", node) as a key and ask if the node type is in the map, if not, panic, if present call the func?

This way you will have one method per type.

Maybe it's naive, and may need a type assertion in each method, but you could split this gigantic piece of code

eval/eval_api.go
```diff
@@ -170,7 +171,8 @@ func AddEvalResult(name, code string) error {
 func EvalString(this any, code string, emptyEnv bool) (object.Object, error) {
 	l := lexer.New(code)
 	p := parser.New(l)
-	program := p.ParseProgram()
+	var program ast.Node
+	program = p.ParseProgram()
```
ccoVeille, line 175: I don't get this change

ast/modify_test.go
ldemailly: == wouldn't be equal as the values (and keys) are pointers to integer

eval/eval_api.go
ldemailly: ParseProgram returns *ast.Statements which happens to be a node (one of the subtype) but when we call Modify() later it returns a Node and we re-assign it (modified copy) into same variable so it has to be of the right type

ast/modify.go
ldemailly: I wouldn't use a string version and keep attaching strings no - now maybe a map can have a type as key, not sure

the code is very mechanical (copy for each type and recurse on each components), a few lines in each case so I don't think the linters alone should be a reason to break it out

ast/modify_test.go
ccoVeille: Oh indeed.

PR #229, opened 2024-09-14
Start of test macros and use + panic in images fix + `for i := num {` fix

ldemailly:
And tests also  adding Mandelbrot sample

grol

### DIFF
--- a/grol_tests/constants.gr
+++ b/grol_tests/constants.gr
@@ -1,31 +1,4 @@
+// - Test constants
 
-/*
-NoErr = macro(expr,a,b) {
-    quote(
-		r = catch(unquote(expr))
-	    if r.err {
-			error("unexpected error:", r.value)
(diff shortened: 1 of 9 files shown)


grol_tests/constants.gr
```diff
@@ -1,31 +1,4 @@
...
-	error("expected error, got ok", r)
-} else {
-	if regexp("attempt to change constant PI", r.value) {
-		println("Got expected error:", r.value)
-	} else {
-		error("unexpected error:", r.value)
-	}
-}
+NoErr("PI should exist", PI, "3.141592")
+IsErr("Can't change a constant", PI++, "attempt to change constant PI")
```
ldemailly, line 4: I think that's pretty nice way of making a bunch of tests, thoughts @ccoVeille ?

It's msg, expression, regexp

PR #237, opened 2024-09-16
Added `run()` and `exec()` functions

### LINKED ISSUE #236: shell like functions
- [ ] exec
- [ ] maybe auto exec from repl mode (though that might not work great with () ?)

ldemailly:
- run() runs to current s.Out and errors out on non 0 exit code
- exec() captures stdout and stderr and err to a map result

Fixes #236


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	fortio.org/safecast v0.1.1
 	fortio.org/sets v1.2.0
 	fortio.org/struct2env v0.4.1
-	fortio.org/terminal v0.8.2
+	fortio.org/terminal v0.8.3-0.20240917004350-924e9d04943f
```
ldemailly, line 11: use 0.9.0 when released before merging this

PR #244, opened 2024-09-21
#204 workaround

MikalaiLappo:
It's a draft PR

1. Makefile should also handle bsd sed

The main bug is probably related to [this workaround](https://github.com/golang/go/issues/28975)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ parser-test:
 TINYGO_STACKS:=-stack-size=40mb
 
 wasm: Makefile *.go */*.go $(GEN) wasm/wasm_exec.js wasm/wasm_exec.html wasm/grol_wasm.html
+	sed -i 's/console.warn("scheduleTimeoutEvent: missed timeout event");/this._scheduledTimeouts.delete(id);\nconsole.warn("scheduleTimeoutEvent: missed timeout event");/'  wasm/wasm_exec.js
 #	GOOS=wasip1 GOARCH=wasm go build -o grol.wasm -trimpath -ldflags="-w -s" -tags "$(GO_BUILD_TAGS)" .
 	GOOS=js GOARCH=wasm $(WASM_GO) build -o wasm/grol.wasm -trimpath -ldflags="-w -s" -tags "$(GO_BUILD_TAGS)" ./wasm
 #	GOOS=wasip1 GOARCH=wasm tinygo build -target=wasi -no-debug -o grol_tiny.wasm -tags "$(GO_BUILD_TAGS)" .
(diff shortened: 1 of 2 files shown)


ldemailly:
so do you see the delete happen once and that stops the flood _and_ the code can further run again after that?

last time the bug/fix/issue was that I hadn't cancel() a timeout context

Makefile
```diff
@@ -36,6 +36,7 @@ parser-test:
 TINYGO_STACKS:=-stack-size=40mb
 
 wasm: Makefile *.go */*.go $(GEN) wasm/wasm_exec.js wasm/wasm_exec.html wasm/grol_wasm.html
+	sed -i 's/console.warn("scheduleTimeoutEvent: missed timeout event");/this._scheduledTimeouts.delete(id);\nconsole.warn("scheduleTimeoutEvent: missed timeout event");/'  wasm/wasm_exec.js
```
ldemailly, line 39: just do the sed in a pipe instead of the copy (where wasm/wasm_exec.js is being created/copied a bit below)

wasm/grol_wasm.html
```diff
@@ -1,3 +1,4 @@
+<!doctype html><html><head><meta charset="utf-8"><title>Grol</title></head>
```
ldemailly, line 1: this is a fragment that is reused mid page, can't have that here

wasm/grol_wasm.html
```diff
@@ -38,12 +39,46 @@
 </style>
 <script src="wasm_exec.js"></script>
 <script>
-    function debounce(func) {
+    /**
+     * A singular task ES6++ mutex
+     * inspired by this https://blog.jcoglan.com/2016/07/12/mutexes-and-javascript/
```
ldemailly, line 44: I was asking on discord about JS mutexes... before being sent the debounce route

does that queue mean if I press the button 10 times (or hit return 10 times in text entry), it'll queue 10 runs? I'm not sure that's better than throwing them away ?

wasm/grol_wasm.html
```diff
@@ -185,3 +219,4 @@
         document.getElementById('input').value = decodeURIComponent(paramValue)
     }
 </script>
+</html>
```
ldemailly, line 222: same, can't be here

ldemailly (changes requested):
thanks a lot for having spent time on this - I'm still wondering what causes the un finished timeout/first timeout

wasm/grol_wasm.html
ldemailly: eg https://github.com/grol-io/web-site/blob/main/butterfly.md?plain=1#L3

ldemailly:
Apologies I didn't try yet, been having too much fun with https://github.com/fortio/terminal - yet thoughts about the feedback? (I think you've been busy too)

PR #272, opened 2024-12-11
`int("")==0`; adding `trim()`, `trim_left()` and `trim_right()`; new `-no-register` option to disable register optimizations

ldemailly:
Despite auto memoization, advent day11 is still too slow - todo: follow up further


extensions/extension.go
```diff
@@ -324,6 +324,8 @@ func createJSONAndEvalFunctions(c *Config) {
 	}
 }
 
+const DefaultTrimSet = " \r\n\t"
```
ccoVeille, line 327:
```suggestion
const defaultTrimSet = " \r\n\t"
```

Unless you plan to reuse it

ldemailly: I figured it would be good to get it documented (thus exposed)

examples/advent_2024_day11.gr
```diff
@@ -0,0 +1,28 @@
...
+			right := int(trim_left(s[l/2:l], "0"))
+			applyRules(left, depth) + applyRules(right, depth)
+		}
+	}
+}
+
+n :=25
+println(applyRules(125,n)+applyRules(17,n))
+
+// Despite auto memoization, this is still too slow even with the new -no-register
```
ldemailly, line 25: follow up with further profiling (already did but nothing stood out) and also maybe check the cache ratio for memoization

PR #369, opened 2025-05-19
(initial) Wasi wasm mode - highlighted need for change in fortio/terminal

ldemailly:
Needs https://github.com/fortio/terminal/pull/121


ldemailly:
will merge this so it gets tagged into fortio and continue in a new branch

PR #381, opened 2025-07-11
Adding +=, -=, *= and /= (compound assignment operators)

### LINKED ISSUE #273: add +=, *= etc (or at least these 2)

geofpwhite:
Fixes #273


eval/eval.go
```diff
@@ -258,7 +265,7 @@ func (s *State) evalInternal(node any) object.Object { //nolint:funlen,gocognit,
 			log.LogVf("eval infix %s", node.DebugString())
 		}
 		// Eval and not evalInternal because we need to unwrap "return".
-		if node.Token.Type() == token.ASSIGN || node.Token.Type() == token.DEFINE {
+		if node.Token.Type() == token.ASSIGN || node.Token.Type() == token.DEFINE || node.Token.Type() == token.SUMASSIGN {
```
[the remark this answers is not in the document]
geofpwhite: What if just check that node.Token.Type() >= token.DEFINE && node.Token.Type() < token.FUNC?

eval/eval.go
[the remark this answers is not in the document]
ldemailly: yeah there shouldn’t be « 37 » there could be whichever token value we subtract 

ps: I am doing family stuff today so will look more tomorrow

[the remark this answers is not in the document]
ldemailly: also I think they could all be variants on assign, just with a subtype inside assign: plain or operator (not sure though, have to see and think more about it)

ps: try to get the Ci to be green meanwhile? (nolint addition is fine in some places)

lexer/lexer.go
```diff
@@ -77,13 +77,20 @@ func (l *Lexer) NextToken() *token.Token {
...
 		return token.ConstantTokenChar(ch)
 	case '+', '-':
-		if nextChar == ch {
+		if nextChar == ch || nextChar == '=' {
 			l.pos++
 			return token.ConstantTokenChar2(ch, nextChar) // increment/decrement
 		}
 		return token.ConstantTokenChar(ch)
-	case '%', '*', ';', ',', '{', '}', '(', ')', '[', ']', '^', '~':
+	case '%', ';', ',', '{', '}', '(', ')', '[', ']', '^', '~':
```
ldemailly, line 85: %= ^= ~= should also work I think - this being said I am curious why is * here and not /

tests/misc.gr
```diff
@@ -102,3 +102,23 @@ Assert("image.delete", image.delete(img) == nil)
 Assert("image.size of deleted image", image.size(img) == nil)
 // extra delete is fine
 Assert("image.delete", image.delete(img) == nil)
+
+i := 4 
+log("before changes", i)
+Assert("i should be 4",i==4)
```
ldemailly, line 108: that test is pretty unnecessary

token/token.go
```diff
@@ -118,6 +118,11 @@ const (
 	LAMBDA // => lambda short cut: `a,b => a+b` alias for `func(a,b) {a+b}`
 	DEFINE // := (force create new variable instead of possible ref to upper scope)
 
+	SUMASSIGN
+	SUBASSIGN
+	PRODASSIGN
+	DIVASSIGN
```
ldemailly, line 124: needs at least a comment to say the order has to be same as PLUS MINUS ASTERISK SLASH

eval/eval.go
```diff
@@ -90,10 +90,17 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	case token.IDENT:
 		id := node.Left.(*ast.Identifier)
 		name := id.Literal()
+		nodeType := node.Type()
+		if nodeType > token.DEFINE && nodeType < token.FUNC {
+			opToEval := nodeType - 37
```
ldemailly, line 95: this `- 37` should be instead `+ token.PLUS - token.SUMASSIGN` no?

ldemailly: or maybe because of overflow issues
```go
nodeType - (token.SUMASSIGN-token.PLUS)
```

geofpwhite: U think of everything. that seems way better. Will make changes accordingly

eval/eval.go
```diff
@@ -90,10 +90,17 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	case token.IDENT:
 		id := node.Left.(*ast.Identifier)
 		name := id.Literal()
+		nodeType := node.Type()
+		if nodeType > token.DEFINE && nodeType < token.FUNC {
```
ldemailly, line 94: >= token.SUMASSIGN && <= token.DIVASSIGN

ie relevant to what this is instead of previous/next ones

eval/eval.go
```diff
@@ -258,7 +265,7 @@ func (s *State) evalInternal(node any) object.Object { //nolint:funlen,gocognit,
 			log.LogVf("eval infix %s", node.DebugString())
 		}
 		// Eval and not evalInternal because we need to unwrap "return".
-		if node.Token.Type() == token.ASSIGN || node.Token.Type() == token.DEFINE {
+		if node.Token.Type() == token.ASSIGN || (node.Token.Type() >= token.DEFINE && node.Token.Type() < token.FUNC) {
```
ldemailly, line 268: same comment and because it's there 2x let's make that change check a function

isAssignement() which is `tokt ==token.ASSIGN  || tokt == token.DEFINE || isCompound(tokt)`
so you can reuse isCompound() above

ldemailly:
bonus:

how much more work to make it work for maps and arrays ?
```
m := {"key": 100}
m.key += 42
m["key"] -= 42

a := 1:4 // 1,2,3
a[1] += 42 // a[1] should be 44
```

eval/eval.go
```diff
@@ -1426,3 +1427,11 @@ func (s *State) stopOutputBuffering() []byte {
 	s.env.PrevOut = nil
 	return output
 }
+
+func isAssignment(node ast.InfixExpression) bool {
```
ldemailly, line 1431: well if you make the signatute the token type instead that would avoid the a bit jarring 
```
if isAssignment(*node) {
```
about as well as save a bunch .Token.Type() here

eval/eval.go
```diff
@@ -1419,3 +1427,11 @@ func (s *State) stopOutputBuffering() []byte {
 	s.env.PrevOut = nil
 	return output
 }
+
+func isAssignment(tok token.Type) bool {
+	return tok == token.ASSIGN || tok == token.DEFINE || isCompound(tok)
+}
+
+func isCompound(tok token.Type) bool {
+	return (tok > token.DEFINE && tok < token.FUNC)
```
ldemailly, line 1436: ahem... as mentioned ... please use >= SUMASSIGN <= SUMDIV and not the unrelated token.FUNC
(which btw also includes a marker gap)

lexer/lexer.go
ldemailly: ic below the / has to handle comments so that's why it was special already

we can add the %= etc later (including &= |= )

tests/misc.gr
```diff
@@ -102,3 +102,22 @@ Assert("image.delete", image.delete(img) == nil)
...
+Assert("i should be 3",i==3)
+i *= i
+Assert("i should be 9",i==9)
+i /= 9.
+Assert("i should be 1.",i==1.)
+i/=2
+Assert("i should be .5",i==.5)
+_str:= "test"
+log("before changes ",_str)
+_str += " + new string"
```
ldemailly, line 120: add a string *= too

eg
```go
str := "ABC"
str *= 3
Assert("multiply string", str == "ABCABCABC")
```

eval/eval.go
```diff
@@ -1419,3 +1427,11 @@ func (s *State) stopOutputBuffering() []byte {
 	s.env.PrevOut = nil
 	return output
 }
+
+func isAssignment(tok token.Type) bool {
+	return tok == token.ASSIGN || tok == token.DEFINE || isCompound(tok)
+}
+
+func isCompound(tok token.Type) bool {
+	return (tok >= token.SUMASSIGN && tok <= token.DIVASSIGN)
```
deitrix, line 1436: The parenthesization isn't necessary here :)

eval/eval.go
```diff
@@ -81,6 +138,13 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	}
 	switch node.Left.Value().Type() {
 	case token.DOT, token.LBRACKET:
+		nodeType := node.Type()
+		if isCompound(nodeType) {
+			opToEval := nodeType - (token.SUMASSIGN - token.PLUS)
```
ldemailly, line 143: maybe isCompound can be/have 

func CompoundOp(type) (op, bool)

so you'd do
```go
if opToEval, isCompound = CompoundOp(nodeType); isCoumpoud {
```
as it's there in 2 places that semi hacky substract

eval/eval.go
```diff
@@ -81,6 +138,13 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
...
+			res, _ := s.compoundAssignNested(node.Left, opToEval, right) // un
```
ldemailly, line 144: // un ?

tests/misc.gr
```diff
@@ -102,3 +102,43 @@ Assert("image.delete", image.delete(img) == nil)
...
+i /= 9.
+Assert("i should be 1.",i==1.)
+i /= 2
+Assert("i should be .5",i==.5)
+i = 1
+log("before changes", i)
+i&=2
+Assert("i should be 0",i==0)
+
+i
```
ldemailly, line 123: lone `i`?

tests/misc.gr
```diff
@@ -102,3 +102,43 @@ Assert("image.delete", image.delete(img) == nil)
...
+i *= i
+Assert("i should be 9",i==9)
+i /= 9.
+Assert("i should be 1.",i==1.)
+i /= 2
+Assert("i should be .5",i==.5)
+i = 1
+log("before changes", i)
+i&=2
+Assert("i should be 0",i==0)
```
ldemailly, line 121: test for |& also use say

5 (0b0101) &= with say 12 (0b1100) to get 4 instead of 0 as more likely to not work vs other ops

and need at least 1 test for |=

eval/eval.go
```diff
@@ -17,6 +17,61 @@ import (
 // See todo in token about publishing all of them.
 var unquoteToken = token.ByType(token.UNQUOTE)
 
+func (s *State) compoundAssignNested(node ast.Node, operator token.Type, value object.Object) (object.Object, *object.Error) {
+	n, ok := node.(*ast.IndexExpression)
+	if !ok {
+		err := s.NewError("assignment to non identifier: " + node.Value().DebugString())
```
ldemailly, line 23: should probably be 
return s.Errorfp("assignment to non identifier:", node.Value().DebugString()) // or maybe .Inspect() not sure

(and assignNested is wrong too, not sure why I have these 
		err := s.NewError("assignment to non identifier: " + node.Value().DebugString())
maybe I wrote that before I wrote Errorfp or forgot - or you can leave it as is and I/we can refactor that seperately)

eval/eval.go
```diff
@@ -81,6 +136,13 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	}
 	switch node.Left.Value().Type() {
 	case token.DOT, token.LBRACKET:
+		nodeType := node.Type()
+		if isCompound(nodeType) {
+			opToEval := nodeType - (token.SUMASSIGN - token.PLUS)
+			res, _ := s.compoundAssignNested(node.Left, opToEval, right) // no need to check error, res is returned either way.
```
ldemailly, line 142: that's likely wrong... it should be
```go
res, oerr := ...
if oerr != nil {
   return oerr
}
```
though the goal is to bubble up in error so maybe only needed in the recursion part 
but I guess in this case there is always something to calculate/return so just change the signature of compoundAssignNested to just return the result (which may be an error obj)

eval/eval.go
```diff
@@ -1149,6 +1218,7 @@ func (s *State) evalStatements(stmts []ast.Node) object.Object {
 			continue
 		}
 		result = s.evalInternal(statement)
+		log.LogVf("result statement %s", result)
```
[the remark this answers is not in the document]
ldemailly: nah it's fine to have logging at verbose or debug level, it helps... debugging

eval/eval.go
```diff
@@ -81,6 +136,13 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	}
 	switch node.Left.Value().Type() {
 	case token.DOT, token.LBRACKET:
+		nodeType := node.Type()
+		if isCompound(nodeType) {
+			opToEval := nodeType - (token.SUMASSIGN - token.PLUS)
+			res, _ := s.compoundAssignNested(node.Left, opToEval, right) // no need to check error, res is returned either way.
+			log.LogVf("%s - right", right)
```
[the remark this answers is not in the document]
ldemailly: it's not the best possible log but... it's fine as well

tests/misc.gr
```diff
@@ -102,3 +102,67 @@ Assert("image.delete", image.delete(img) == nil)
 Assert("image.size of deleted image", image.size(img) == nil)
 // extra delete is fine
 Assert("image.delete", image.delete(img) == nil)
+
+i := 4
```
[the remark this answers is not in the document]
ldemailly: he is right about that one, set your vscode to remove trailing spaces
Screenshot 2025-07-18 at 2 12 49 PM

tests/misc.gr
```diff
@@ -102,3 +102,67 @@ Assert("image.delete", image.delete(img) == nil)
...
+m[[1,2]]*=3 // -> "abcabcabc"
+Assert("m[[1,2]] should be 'abcabcabc'",m[[1,2]]=="abcabcabc")
+
+m[[1,2]]="a"
+m[[1,2]]+="b"
+Assert("m[[1,2]] should be 'ab'",m[[1,2]]=="ab")
+
+ x = {"a":{"b":{"c":1}}}
+log("before changes x = {a:{b:{c:1}}}",x) 
+ x.a.b.c|=12
```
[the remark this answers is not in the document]
ldemailly: minor but yeah kinda, leading extra spaces ?

eval/eval.go
ldemailly: moving to #385

eval/eval.go
```diff
@@ -81,6 +137,13 @@ func (s *State) evalAssignment(right object.Object, node *ast.InfixExpression) o
 	}
 	switch node.Left.Value().Type() {
 	case token.DOT, token.LBRACKET:
+		nodeType := node.Type()
+		if isCompound(nodeType) {
+			opToEval := nodeType - (token.SUMASSIGN - token.PLUS)
+			res, _ := s.compoundAssignNested(node.Left, opToEval, right) // no need to check error, res is returned either way.
```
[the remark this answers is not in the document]
ldemailly: I do think you should either use or rather remove the err return if not used/useful
(had mentioned at the end of https://github.com/grol-io/grol/pull/381#discussion_r2216830494)

eval/eval.go
```diff
@@ -17,6 +17,62 @@ import (
...
+	var identifier string
+	if id, ok := left.(*ast.Identifier); ok {
+		identifier = id.Literal()
+		baseObj, ok := s.env.Get(identifier)
+		if !ok {
+			err := s.NewError("identifier not found: " + identifier)
+			return err, &err
+		}
+		base = object.Value(baseObj)
+	} else {
```
ldemailly, line 40: identifier isn't set in the else and yet used line 62?

ldemailly:
I forgot to formally "approve" (because it's not (yet) mandatory in this repo...) but I approve :) [with #387 as followup]

PR #443, opened 2025-11-07
Add float() and allow spaces in string input for int()

### LINKED ISSUE #442: consider a float() to convert any to float (in particular strings to float64)
so 

float("0x1.921FB54442D18p+1") == PI 

would be true for instance (as we don't support that format in the parser; optionally we could also... support it in the parser too)

ldemailly:
- Fixes #442 (in 1a5a2d3df8f15cf0e9084b3fe043fe8e6fdc1e17)
- Linter related changes

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -133,7 +133,7 @@ type Node interface {
 	PrettyPrint(ps *PrintState) *PrintState
 }
 
-// Base is common to all nodes that have a token and avoids repeating the same TokenLiteral() methods.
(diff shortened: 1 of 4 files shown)


ldemailly:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/grol-io/grol/pull/443#pullrequestreview-3435839619)

ldemailly:
what a disaster of noise this copilot tiny updates... can't batch merge ...

PR #475, opened 2026-01-07
Add support for single quote rune literal as int

ldemailly:


ldemailly:
still a bit ugly but... let's go (need it in vm to make the assembly smarter)

